PR #8, opened 2018-04-24
Reference folder has been updated with latest changes

ayxos:
AppNexus repo has been updated: https://github.com/appnexus/cmp


tomeightyeight:
Can we get this merged or much better maybe just make reference to the AppNexus repo?? I wrongly assumed the reference implementation included here was up to date and the primary source for that code base.

PR #10, opened 2018-04-30
Corrections to example data

simontrasler:
1. The stated timestamp and binary representation did not match.
2. The example consent string did not match the preceding table.
3. The sections on Language reference uppercase and give an example in lowercase.

### DIFF
--- a/Consent string and vendor list formats v1.1 Final.md
+++ b/Consent string and vendor list formats v1.1 Final.md
@@ -603,7 +603,7 @@ Example consent string field values for the case:
     <td>SingleVendorId[0]</td>
     <td>9</td>
     <td>VendorId=9 has No Consent (opposite of Default Consent)</td>
-    <td>00000000000001001</td>
+    <td>0000000000001001</td>
   </tr>
   <tr>
     <td>base64url-encoded consent string value</td>
--- a/Consent string and vendor list formats v1.1 Final.md
+++ b/Consent string and vendor list formats v1.1 Final.md
@@ -609,7 +609,7 @@ Example consent string field values for the case:
     <td>base64url-encoded consent string value</td>
     <td></td>
     <td></td>
-    <td>BOEFEAyOEFEAyAHABDENAI4AAAB9vABAAS</td>
+    <td>BOEFEAyOEFEAyAHABDENAI4AAAB9vABAASA</td>
   </tr>
 </table>
 
--- a/Consent string and vendor list formats v1.1 Final.md
+++ b/Consent string and vendor list formats v1.1 Final.md
@@ -524,12 +524,12 @@ Example consent string field values for the case:
   <tr>
     <td>Created </td>
     <td>15100811449 </td>
-    <td>2017-11-07T18:59:04.9Z</td>
+    <td>2017-11-07T19:15:55.4Z</td>
     <td>001110000100000101000100000000110010</td>
   </tr>
   <tr>
     <td>LastUpdated</td>
-    <td>15100811449</td>
+    <td>15100821554</td>
     <td>2017-11-07T18:59:04.9Z</td>
     <td>001110000100000101000100000000110010</td>
   </tr>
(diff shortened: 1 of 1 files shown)


Consent string and vendor list formats v1.1 Final.md
```diff
@@ -524,12 +524,12 @@ Example consent string field values for the case:
   <tr>
     <td>Created </td>
     <td>15100811449 </td>
-    <td>2017-11-07T18:59:04.9Z</td>
+    <td>2017-11-07T19:15:55.4Z</td>
```
dhull: There is still something wrong here, because this field is supposed to be deci-seconds, so for the date `2017-11-07T19:15:55.4Z` the last digit of the integer value in column 2 should be `4`. According to my calculation, the translation of `2017-11-07T19:15:55.4Z` into epoch decisecond value `15100811449` is correct, The error is that the binary representation of 15100811449 is `001110 000100 000101 000001 100010 111001`.

Also, you are correct that the Base64 encoding of the example is incorrect.

simontrasler: Gah, I forgot to update the decimal value, doing now.

dhull:
My encoder is only half-written, but I can confirm that the conversion between time, epoch deciseconds, and binary in the Created and LastUpdated fields are now correct, and that the conversion of of the fields up through ConsentLanguage into Base64 is correct.

Consent string and vendor list formats v1.1 Final.md
```diff
@@ -609,7 +609,7 @@ Example consent string field values for the case:
     <td>base64url-encoded consent string value</td>
     <td></td>
     <td></td>
-    <td>BOJObISOJObISAABAAENAA4AAAAAoAAA</td>
+    <td>BOEFEAyOEFEAyAHABDENAI4AAAB9vABAAS</td>
```
dhull: I believe this should probably be `BOEFEAyOEFEAyAHABDENAI4AAAB9vABAASA` (with an additional `A` on the end).

Although the spec does not explicitly say that the vendor consent string should be padded out to an integral number of bytes, it does say that for the global vendor consent cookie and the publisher purposes consent string, so I would guess that for consistency it should be done here as well.

The example consent string ends up being 203 bits long. We need to pad that out with 5 `0` bits to bring it to an even 26 bytes. When we Base64 encode that, we get `BOEFEAyOEFEAyAHABDENAI4AAAB9vABAASA=`. If we then drop the trailing `=` we get `BOEFEAyOEFEAyAHABDENAI4AAAB9vABAASA`.

Other than that, the value looks correct to me.

Consent string and vendor list formats v1.1 Final.md
```diff
@@ -387,7 +387,7 @@ user?s consent action timing. Javascript: Math.round((new Date()).getTime()/100)
     <td>12 bits
 (108-119)</td>
     <td>Two-letter ISO639-1 language code that CMP asked for consent in</td>
-    <td>Each letter should be encoded as 6 bits, a=0..z=25 . This will result in the base64url-encoded bytes spelling out the language code (in uppercase).</td>
+    <td>Each letter should be encoded as 6 bits, A=0..Z=25 . This will result in the base64url-encoded bytes spelling out the language code (in uppercase).</td>
```
dhull: I'm not sure this change is needed. The ISO 639-1 codes are lowercase, and the previous version of the description made it clear that `a` mapped to 0 and `z` mapped to 25. But it's OK either way.

Consent string and vendor list formats v1.1 Final.md
simontrasler: Point taken, agreed, changing now.

Consent string and vendor list formats v1.1 Final.md
simontrasler: Ah I see. I found the documentation confusing. It's possible to use lowercase in the base64 encoding so I wasn't clear why this needed to be uppercased.

dhull:
This pull request should be merged. It contains important corrections to the example vendor consent string.

Consent string and vendor list formats v1.1 Final.md
```diff
@@ -609,7 +609,7 @@ Example consent string field values for the case:
     <td>base64url-encoded consent string value</td>
     <td></td>
     <td></td>
-    <td>BOJObISOJObISAABAAENAA4AAAAAoAAA</td>
```
lookAtThat: There are too many bits in the **VendorID=9** field above. Theres 17 bits, it should be 16.

PR #13, opened 2018-05-02
iOS/Android Wrappers

ayxos:
- iOS wrapper
- Android wrapper
- custom reference for mobile environment


Kowshickkarthick:
I can see most of the files that show up as change or new in path `/reference/src` are already part of the `InteractiveAdvertisingBureau:master` branch. 
Can we rebase the`smaato:wrappers` to the latest of  `InteractiveAdvertisingBureau:master`  and see if this goes away.
For example this file is present in InteractiveAdvertisingBureau:master https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/reference/src/docs/assets/vendorlist.json but it shows up as new in the PR.

That way we have a cleaner set of changes to look at. And we do not unknowingly break anything for the Web CMP JS reference implementation.

In-App Reference/Android/app/src/main/java/com/smaato/soma/cmpconsenttooldemoapp/cmpconsenttool/model/SubjectToGdpr.java
```diff
@@ -0,0 +1,31 @@
+package com.smaato.soma.cmpconsenttooldemoapp.cmpconsenttool.model;
+
+/**
+ * Enum that indicates:
+ * 'CMPGDPRDisabled' – value 0, not subject to GDPR,
+ * 'CMPGDPREnabled'  – value 1, subject to GDPR,
+ * 'CMPGDPRUnknown'  - value -1, unset.
+ */
+public enum SubjectToGdpr {
+    CMPGDPRUnknown("-1"), CMPGDPRDisabled("0"), CMPGDPREnabled("1");
```
Kowshickkarthick: The value for `CMPGDPRUnknown` is `Nil` as per the Spec and not `-1`

Here is how the values are defined in the draft spec.
```String 1 - (subject to GDPR), 0 - (not subject to GDPR), Nil - undetermined (default before initialization)```

Kowshickkarthick:
@jenniferIAB , I am not sure on what the IAB's policy is for file license and links to hosted files in this GitHub space. But just noting that 

1) The files in path `~/In-App Reference/iOS` contain  `Copyright © 2018 Smaato. All rights reserved.`

2)  The WebPage that the implementation  references is in path `https://demofiles.smaato.com/cmp/index.html`

ppuviarasu:
In the consent UI for IOS, the navigation of the consent views in not adhering to the Human Interface Guidelines mentioned by Apple... I worry that would be a problem...

ppuviarasu:
Vendor choices made on each tab...  is it the same list for entire consent or for individual line items.
Rite now the choice made under each section is applied to all the sections

PR #14, opened 2018-05-02
Replaced a few question marks with apostrophes.

eliotpearson:
I believe these are typos. Please decline the PR if question marks were intended.


eliotpearson:
Closing this PR. Looks like this section has been removed.

PR #16, opened 2018-05-03
Fix typo and discrepancies in vendor consent example

izhamoidsin:
`15100811449` is equivalent to `2017-11-07T18:59:04.9Z` but in binary form it is represented as `001110000100000101000001100010111001`, but not `001110000100000101000100000000110010`

VendorId=9 is encoded `00000000000001001` but according to the spec it should consist of 16 bits. Removing extra 0 in the head of the number.

Full consent string in binary format is `0000010011100001000001010001000000001100100011100001000001010001000000001100100000000001110000000000010000110001000011010000000010001110000000000000000000000000011111011011110000000000010000000000000100100000` (assembled from values mentioned in example table and padded to the nearest multiple 8 bits). Transforming to bytes and applying Base64 URL encoding gives the value of `BOEFBi5OEFBi5AHABDENAI4AAAB9vABAASA` not `BOJObISOJObISAABAAENAA4AAAAAoAAA` (validated via )


mlyko:
Are you sure about correctness of the base64 encoded content string which you've just fixed in the corresponding commit?
I get BOEFBi5OEFBi5AHABDENAI4AAAB9vABAASA with GDPR Consent Cookie workshop:
http://gdpr-demo.labs.quantcast.com/user-examples/cookie-workshop.html

izhamoidsin:
Hi @mlyko 
Good catch. It seems there is a typo in typo fix pull request. 
`BOEFBi5OEFBi5AHABDENAI4AAAB9vABAASA` is the valid one. Verified both through http://gdpr-demo.labs.quantcast.com/user-examples/cookie-workshop.html and http://acdn.origin.appnexus.net/cmp/docs/#/tools/vendor-cookie-decoder.

I have fixed the binary value for dates but kept the old one `001110000100000101000100000000110010` (equal to `2017-11-07T19:15:55.4Z`) compiled the bit string and encoding it.

chrispaterson:
Thanks guys!

PR #22, opened 2018-05-10
Added LR Golang Implementation of Vendor Consent String Spec

adayNU:
We implemented a Golang version of the Vendor Consent String Spec.


adayNU:
cc: @jenniferIAB

dbemiller:
d'oh.... it looks like we both did some work in parallel :x.

I also put together a Golang lib with support for vendor consent strings, as well as the global vendor list. Probably worth collaborating? Not much sense in building two different ones.

https://github.com/prebid/go-gdpr

PR #29, opened 2018-05-22
Add list of unofficial third-party SDKs for other languages

cirla:
Add list of unofficial third-party SDKs for other languages. Extends #22


jenniferIAB:
Merged. Thank you for helping expand our references!

PR #33, opened 2018-05-30
Cleanup third party SDK section

cirla:
1. Remove duplicate link
2. Make origins clear and visible when there is more than one implementation for a given language.


chrispaterson:
Good catch, thanks!

PR #56, opened 2018-07-11
Changing specs to add missing header consentLanguage to metadata from…

jdelhommeau:
Adding missing header consentLanguage to metadata from getVendorConsents method.

ConsentLanguage header present in the consentData as per consentData specs (https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/Consent%20string%20and%20vendor%20list%20formats%20v1.1%20Final.md) was missing from the metadata field present in response from getVendorConsents method.
To be more coherent across the specs, this value should also be present in the metadata field from the getVendorConsents result.

### DIFF
--- a/CMP JS API v1.1 Final.md
+++ b/CMP JS API v1.1 Final.md
@@ -250,9 +250,11 @@ where *vendorId* and *purposeId* are the keys and *consentBoolean *are the value
 
 6. Consent Screen
 
-7. Vendor List Version
+7. Consent Language
 
-8. Publisher Purposes Version (for the *PublisherConsent* metadata only)
+8. Vendor List Version
+
+9. Publisher Purposes Version (for the *PublisherConsent* metadata only)
 
 ### VendorConsentData <a name="vendorconsentdata"></a>
 


jawadst:
I agree with the spirit of the change but that's a breaking change and should be scheduled for 1.2 rather than be a modification to the 1.1 that is already out.

For instance, here is the list of fields in the metadata for the JS implementation: https://github.com/InteractiveAdvertisingBureau/Consent-String-SDK-JS/blob/master/src/utils/definitions.js#L18

jdelhommeau:
Thanks @jawadst indeed makes sense to add it for 1.2. Do you know when will this be ready? How do we port those changes so they can be ready for 1.2 ?

chrispaterson:
agree with @jawadst , @jdelhommeau let me figure out a way to capture these ideas for 1.2.

jdelhommeau:
@chrispaterson  and @jawadst  It seems like this issue has raised lot of concerns lately. David Dabbs from Conversant came up with another approach, where rather than changing existing field (making it a beaking change) we could add a new one, containing all the information, + adding all decoded fields. 
That way we can publish it right away without breaking existing implementation.

Does that feel reasonable?

jawadst:
@jdelhommeau Making it non-breaking is indeed much better but still makes it a change that requires releasing a new version so that all CMPs can update.

PR #57, opened 2018-07-11
adding decoded header info to getVendorConsents returned object

jdelhommeau:
adding decoded header info to getVendorConsents returned object

Since getVendorConsents goal is to return decoded information from the consentString, it makes sense to provide decoded headers value as well since those could be useful for vendors that whish to exploit consentData information client side.


jawadst:
I agree with the idea but that should be a proposal for 1.2 and not a change to the 1.1 that is already out.

chrispaterson:
I agree with @jawadst on this one.  Let's move this to incorporate in the 1.2 spec.

PR #62, opened 2018-07-20
Encoding warning

jdelhommeau:
Adding warning regarding encoding in base64url for consent string as currently lot of confusion due to people trying to convert the consentString from a **character string** to a **base64** encoding instead of a **bytes string** to a **base64url** encoding resulting in inconsistent behavior


chrispaterson:
I think this is a little bit too much.  The above links specify that it is base64_url encoded.

Issue #63, opened 2018-07-20
Missing 0-bits in example


bundeskanzler4711:
Hi!
In the "Example Vendor Consent String" (part of "Consent string and vendor list formats v1.1 Final.md") there are some 0-bits missing at the end of the cookie value field bits.
These five fill-bits have to be added in order to pad the binary bits at the end with zeroes to the nearest multiple of 8 bits.
If this information is missing in the example it stays unclear which binary bits leads to the correct base64 string.

PR #64, opened 2018-07-20
Fix missing 0 bits

bundeskanzler4711:
This PR tries to resolve the issue #63:
- add fill-bits in the example table
- add a byte-oriented overview of the complete binary bits
- move the base64 string to the new explanation part


chrispaterson:
What are your thoughts on this @a2intl ?  DO you think the extra detail here is helpful?  I think you wrote the spec initially?

a2intl:
Thanks @chrispaterson and @bundeskanzler4711 , this additional detail and clarification is helpful.

PR #66, opened 2018-08-11
Spec does not match what is actually being done

pdehaye:
The spec specifies that purpose 2 is tied to "Personalisation", but the description presented in the spec does not match [what is being done in practice now](https://vendorlist.consensu.org/vendorlist.json), from [day 1 of the `vendorlist.json`](https://vendorlist.consensu.org/v-1/vendorlist.json). Indeed, the spec mentions personalisation of advertisements only, while the actual `vendorlist.json` file, which contains what will in practice be displayed to the users, says that this is for personalisation of advertisements AND content. 

Note that while this is "just" a Github issue, this file ([timestamped](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/commit/a32574941ce201708e30e78702278efe1ce6cd59)) is actually [referred to on the IAB website as the official spec](http://advertisingconsent.eu/publishers/) (see "CONSENT STRING AND GLOBAL VENDOR LIST FORMAT SPEC V1.1​"). Depending on who refers to this file and who refers to the actual `json`, this will have serious consequences on the validity of the consents traveling the ecosystem based on purpose 2.

Given that this is precisely the topic of several ongoing lawsuits against advertisers, and that long term control over this data is directly tied to significant power dynamics between the different actors (content producers vs advertisers), it is a pretty wild discrepancy to have let slip through.

### DIFF
--- a/Consent string and vendor list formats v1.1 Final.md
+++ b/Consent string and vendor list formats v1.1 Final.md
@@ -239,8 +239,8 @@ Features are informational and should be shown on the UI as a data use by that v
   </tr>
   <tr>
     <td>2</td>
-    <td>Personalisation</td>
-    <td>The collection and processing of information about user of a site to subsequently personalize advertising for them in other contexts, i.e. on other sites or apps, over time. Typically, the content of the site or app is used to make inferences about user interests, which inform future selections.</td>
+    <td>Ad and content personalisation</td>
+    <td>The collection and processing of information about user of a site to subsequently personalize advertising or content for them in other contexts, i.e. on other sites or apps, over time. Typically, the content of the site or app is used to make inferences about user interests, which inform future selections of advertising or content.</td>
   </tr>
   <tr>
     <td>3</td>


chrispaterson:
I've asked some policy folks to review this.

chrispaterson:
I see no discrepancy.  I see no documents referring to "Ad and content personalisation" but only "Personalisation" as it is in the official policy documents http://www.iabeurope.eu/tcfdocuments/documents/legal/currenttcfpolicyFINAL.pdf the published vendorlist json https://vendorlist.consensu.org/vendorlist.json or the policy page on github which you are submitting a pull request for.

chrispaterson:
If you are attempting to make a legal argument, I would notify the IAB via email.

pdehaye:
Modulo some formatting HTML/json, there are discrepancies between the following two (hyperlinked copy/pastes): 

A. [here](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/commit/a32574941ce201708e30e78702278efe1ce6cd59#diff-89743c721a4fbc6c8b502775f30b5b23R253): 
```
    <td>Personalisation</td>
    <td>The collection and processing of information about user of a site to subsequently personalize advertising for them in other contexts, i.e. on other sites or apps, over time. Typically, the content of the site or app is used to make inferences about user interests, which inform future selections.</td>
``` 
B. [here](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/75a7c0d796c96b8b3878e8a4e23aea0bd99ba63a/reference/src/docs/assets/vendorlist.json#L13):
```
{"id":2,"name":"Personalisation","description":"The collection and processing of information about your use of this service to subsequently personalise advertising and/or content for you in other contexts, such as on other websites or apps, over time. Typically, the content of the site or app is used to make inferences about your interests, which inform future selection of advertising and/or content."}
```

Some of the discrepancies are explained by the fact that in one case that particular purpose is presented to the data subject (B: *about your use of the service*), while in the other it is to someone else than the data subject (presumably publishers in A: *about user of a site*). I think that type of discrepancy is ok. 

I note however a *very* significant discrepancy in the stated final purpose: "personalize advertising" (A), "personalise advertising and/or content" (B). Again, this concerns content that is inside the GitHub repo (I have hyperlinked the exact line). 

**It is a statement of fact that there is a discrepancy, separately of any legal argument.**

I am not building and will not build a legal argument here, but nevertheless observe that A is what is [referred to on the IAB website as the official spec (on the publisher facing part of the site!)](http://advertisingconsent.eu/publishers/), while B is what is actually used operationally.

PR #70, opened 2018-08-31
Fixing documentation -- changing misinterpreted apostrophe back to an…

chrispaterson:
… actual apostrophe


chrispaterson:
No longer a problem

PR #73, opened 2018-09-25
Sample Stub code edits

chrispaterson:
Adding try catch around JSON parse to prevent postMessage errors for junk.  Adding curley braces around if statements to prevent minification errors.  Aliasing, window and document to variables for minification.  Also removed reference to attachEvent since that is an IE only method that predates postMessage.


chrispaterson:
Also realized a bug on ie11 where iframe.style is a read-only property and ie11 is unhappy with that.  I just made a change to iframe.style.cssText instead.

dav-m85:
Can we get this merged please ? The current stub implementation fails in real life situations, as it catches all postMessage trafic and fails to decode it obviously. Thanks.

PR #75, opened 2018-10-12
Update Consent string and vendor list formats v1.1 Final.md

dmdabbs:
Matthias has indicated in several meetings that there is no Policy underlying this limitation the tech spec places on use of the Framework web domain. Therefore propose that we remove this limitation from the technical specification.

### DIFF
--- a/Consent string and vendor list formats v1.1 Final.md
+++ b/Consent string and vendor list formats v1.1 Final.md
@@ -603,7 +603,7 @@ The calculated value for this example is:
 
 Under this proposal, there are two types of publisher-specific consent: service-specific vendor consent, and a publisher's purposes consent for its own data use (if needed). For example, a publisher that wants to set a frequency-capping first-party cookie should request Publisher Purposes Consent for Purpose #1 "Accessing a Device".
 
-Service-specific vendor consent can be implemented (optionally) by a CMP using the same Vendor Consent String Format (as detailed above) with the Host, Path, and/or Cookie Name referencing a first-party cookie (or service-shared 3rd-party cookie).  If a CMP allows a subset of vendors to be configured by a publisher, storage of consent in a service-specific cookie is necessary to avoid clearing previously-given consent (on non-included vendors) on the global cookie.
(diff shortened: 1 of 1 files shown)


Consent string and vendor list formats v1.1 Final.md
```diff
@@ -603,7 +603,7 @@ The calculated value for this example is:
 
 Under this proposal, there are two types of publisher-specific consent: service-specific vendor consent, and a publisher's purposes consent for its own data use (if needed). For example, a publisher that wants to set a frequency-capping first-party cookie should request Publisher Purposes Consent for Purpose #1 "Accessing a Device".
 
-Service-specific vendor consent can be implemented (optionally) by a CMP using the same Vendor Consent String Format (as detailed above) with the Host, Path, and/or Cookie Name referencing a first-party cookie (or service-shared 3rd-party cookie).  If a CMP allows a subset of vendors to be configured by a publisher, storage of consent in a service-specific cookie is necessary to avoid clearing previously-given consent (on non-included vendors) on the global cookie.
+Service-specific vendor consent can be implemented (optionally) by a CMP using the same Vendor Consent String Format (as detailed above) with the Host, Path, and/or Cookie Name referencing a first-party cookie (or service-shared 3rd-party cookie).
```
a2intl: There is not the technical support for the removal of this restriction in the spec, yet, even if this is not supported by policy (it's a technical restriction, not a policy restriction). If we do want to support vendor sublists for the global vendor cookie, we would first need to spec out how the masking of a service-specific vendor sublist on the global vendor list should be implemented (i.e. it should be masked on the API return values, but not in the cookie value itself), and we'd also need to add a separate "vendors seen" cookie that a CMP could use to determine whether all the vendors _they_ need consent for have been shown to the user for consent yet. @chrispaterson

dmdabbs: Agree @a2intl. The masking scheme you're referencing is the 'Rendered Vendors" proposal. As part of today's discussion and Matthias's comments, Jennifer suggested that I submit a PR for this removal from the spec document to nudge things forward.

chrispaterson:
I believe this is the policy going forward for 1.1.  We may wish to revisit this in 2.0, but for now, it stays.

PR #77, opened 2018-11-28
Updates Reference CMP Code - uses example CMP_ID

Grantimus9:
The AppNexus reference implementation used CMP_ID=1 as an example value, expecting CMP implementations to replace the value with their own IAB-assigned CMP ID. We've updated the AppNexus implementation to now use CMP ID 0 as an example, and added a note indicating that future implementers of the code should register for a CMP ID. 

This PR brings the IAB reference code up-to-date in this regard. 

For reference, this is merged in the CMP reference implementation: 
image.

### DIFF
--- a/reference/src/lib/init.js
+++ b/reference/src/lib/init.js
@@ -9,7 +9,11 @@ import pack from '../../package.json';
 import config from './config';
 
 const CMP_VERSION = 1;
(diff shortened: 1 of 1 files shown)


Grantimus9:
🎉

PR #107, opened 2019-11-14
Update global cookie support

jenniferIAB:
In order to support global cookie with current browser changes. Chrome 80, due to be launched in February 2020, requires cookie attributes 'SameSite=None' and 'Secure' to be set for 3rd party cookies to work, as well as any access to cookies to be over HTTPS.


chrispaterson:
Looks good!

PR #111, opened 2019-12-06

wittjill:
update macros from lower case to upper case


chrispaterson:
Looks good

Issue #113, opened 2019-12-06
Is the appendix missing?


mdenison:
We are currently implementing a small SDK for Java to support TCF 2.0, especially for the TC String. When looking at https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20Consent%20string%20and%20vendor%20list%20formats%20v2.md#the-core-string it states "Example values are provided in the appendix." . 

Am I overseeing the appendix? Or is it perhaps actually missing? 

Best thanks in advance.

chrispaterson:
I believe this is carry-over text from the previous version.  I think the intention was to put these types of examples in the Implementation Guidelines: https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/TCF-Implementation-Guidelines.md 

But they aren't there either...

chrispaterson:
I think static examples are not going to be super useful.  If you want to generate some encodings you can use this: https://www.iabtcf.com/#/encode

PR #118, opened 2019-12-11

wittjill:
remove reference to example encoding appendix


chrispaterson:
Looks good, this addresses #113

PR #122, opened 2020-01-08
Consistent camel-casing for Optins

nathan-lapinski:
Fixing a potential type. Please ignore if this is a non-issue.

Should these be `Optins` or `OptIns`?
`IABTCF_SpecialFeaturesOptIns` uses `OptIns`.


wittjill:
while good practice, the previous is already out in the ecosystem and potentially implemented. Can't make this trivial change when it could have breaking impacts.

PR #126, opened 2020-01-13
Proposed changes in spec doc

Leedehai:
#### Change addEventListener/removeEventListener interface

[window.postmessage()](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) cannot [serialize](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) a function object, so the
original spec for removeEventListener does not work cross-frame.
To handle this, let CMP maintain a internal map of registered
listeners and their IDs.

- TCData layout: add `listenerId`
- removeEventListener `parameter` param: `callback` -> `listenerId`

#### Clarify addEventListener callback invocation time

Emphasize that the callback should be invoked with the current TC
String upon registration, so that it makes sense to recommend
"calling scripts register a listener function via addEventListener
instead of getTCData" in the `getTCData` section.

#### Remove SafeFrame proxy communications.

Remove descriptions on safeFrame proxy since it is incompatible
with caller sync expectations.

### DIFF
--- a/TCFv2/IAB Tech Lab - CMP API v2.md
+++ b/TCFv2/IAB Tech Lab - CMP API v2.md
@@ -262,7 +262,7 @@ The CMP will, in most cases, invoke the callback when  either the `'tcloaded'` O
 
 The callback shall be invoked with `false` as the argument for the `success` parameter if the callback could not be registered as a listener for any reason.
 
-**Note:** The `addEventListener` callback shall be immediately called upon registration with the current TC data, then subsequently called again on any TC String changes unless it is removed via `removeEventListener`.
+> **Note**: The `addEventListener` callback shall be immediately called upon registration with the current TC data, even if the CMP status is `loading` and the CMP has incomplete TC Data, so that the calling script may have access to its registered `listenerId`. Furthermore, on every TC String change the callback shall be called unless it is removed via `removeEventListener`.
 ______
 
 #### `removeEventListener`
--- a/TCFv2/IAB Tech Lab - CMP API v2.md
+++ b/TCFv2/IAB Tech Lab - CMP API v2.md
@@ -46,8 +46,7 @@
     - [Requirements for the CMP “stub” API script](#requirements-for-the-cmp-stub-api-script)
     - [Is there a sample CMP “stub” API script?](#is-there-a-sample-cmp-stub-api-script)
   - [How can vendors that use iframes call the CMP API from an iframe?](#how-can-vendors-that-use-iframes-call-the-cmp-api-from-an-iframe)
-    - [Via safeFrames](#via-safeframes)
-    - [Without safeFrames, using postMessage](#without-safeframes-using-postmessage)
+    - [Using postmessage](#using-postmessage)
     - [Is there a sample iframe script call to the CMP API?](#is-there-a-sample-iframe-script-call-to-the-cmp-api)
   - [From where will the API retrieve the TC string?](#from-where-will-the-api-retrieve-the-tc-string)
     - [How will the API prioritize the service-specific and the global configurations?](#how-will-the-api-prioritize-the-service-specific-and-the-global-configurations)
@@ -140,7 +139,7 @@ Every consent manager MUST provide the following API function:
 
 The function `__tcfapi` **must always be a function** and cannot be any other type, even if only temporarily on initialization – the API must be able to handle calls at all times.
 
-Secondarily, CMPs must provide a proxy for postMessage events targeted to the `__tcfapi` interface sent from within nested iframes. See [the section on iframes](#how-can-vendors-that-use-iframes-call-the-cmp-api-from-an-iframe) for information on working with IAB SafeFrames.
+Secondarily, CMPs must provide a proxy for postMessage events targeted to the `__tcfapi` interface sent from within nested iframes. See [the section on iframes](#how-can-vendors-that-use-iframes-call-the-cmp-api-from-an-iframe) for information.
 
 ### What required API commands must a CMP support?
 
@@ -237,7 +236,7 @@ const callback = (tcData, success) => {
 
       }
 
-    }, callback);
+    }, tcData.listenerId);
 
 
   } else {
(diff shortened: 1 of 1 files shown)


TCFv2/IAB Tech Lab - CMP API v2.md
```diff
@@ -388,6 +386,14 @@ TCData = {
    */
   eventStatus: 'string',
 
+  /**
+   * If this TCData is sent to the callback of addEventListener: number,
+   * the unique ID assigned by the CMP to the listener function registered
+   * via addEventListener.
+   * Others: undefined.
```
chrispaterson: I think the listenerId should be a string and it should be empty if none exists.

chrispaterson: Likely this will be used as a key in a map

TCFv2/IAB Tech Lab - CMP API v2.md
```diff
@@ -263,8 +262,7 @@ The CMP will, in most cases, invoke the callback when  either the `'tcloaded'` O
 
 The callback shall be invoked with `false` as the argument for the `success` parameter if the callback could not be registered as a listener for any reason.
 
-**Note:** Unlike the other API commands, the `addEventListener` callback may be called as many times as the TC String is changed — callback functions should be defensive and remove themselves as listeners if this behavior is not desired via `removeEventListener`.
-
+**Note:** The `addEventListener` callback, upon registration, should be immediately called with the current TC String, then subsequently called on any TC String changes until it is removed as listeners via `removeEventListener`.
```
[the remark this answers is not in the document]
Leedehai: Makes sense, will update.

TCFv2/IAB Tech Lab - CMP API v2.md
Leedehai: Could you please elaborate on why strings being versatile would help?
Additionally, requiring it to be an empty string if none exists means the TCData object must contain the field name no matter what (if we allow it to be `undefined`, TCData can leave it out), adding transmission overhead, albeit very small.

TCFv2/IAB Tech Lab - CMP API v2.md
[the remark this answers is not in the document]
mlb7687: If that is the case, why can't a CMP callback immediately with a TCData that reflects a CmpStatus/EventStatus that reflects "invalid" TCData instead?
We discussed how the notion of "valid" TCData isn't 100% consistent depending on interpretation/legit interest vs consent/etc, so gating the behavior on "valid" TCData seems like it leaves this behavior ambiguous as well, no?

TCFv2/IAB Tech Lab - CMP API v2.md
mlb7687: +1, isn't clear to me why a string is superior here, but open to being convinced.
A CMP could represent as strings like '1234' internally and just cast to numbers for TCData if that allows them more flexibility with their implementation.

TCFv2/IAB Tech Lab - CMP API v2.md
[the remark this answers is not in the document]
chrispaterson: I see your point.  I think `cmpStatus = 'loading'` is the status I'm looking for here.   So I think we can omit the 'or as soon as'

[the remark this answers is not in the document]
Leedehai: It might be better to add "even if the CMP status is not `loaded`", to make the requirement explicit, quashing any doubt.
**Note:** The `addEventListener` callback shall be immediately called upon registration with the current TC data, even if the CMP status is not `loaded`, then subsequently called again on any TC String changes unless it is removed via `removeEventListener`.

TCFv2/IAB Tech Lab - CMP API v2.md
chrispaterson: It doesn't matter that much whether it's a string or a number.  The only thing that holds me up a little on the number is that "number" properties on objects can't be 'dot' accessed – perhaps a minor limitation.  It's not a hill I want to die on ;-)

@Leedehai, I agree with you on the undefined, probably better to omit (undefined) if not necessary.  But to be consistent with the format of the other annotations we should not use a union type to describe it and instead make a note in the comment above it that it may be undefined.   Conversely, we could add the union type to gdprApplies which can also be undefined.  I just think they should be consistent.

Leedehai: Will annotate `gdprApplies` with a union type, thinking it might be easier to read even without the comments or convert it to a [.d.ts](https://www.typescriptlang.org/docs/handbook/advanced-types.html#union-types) file.

As for the listenerId, I still prefers a numeric type because it's in line with other "id" fields in the spec. But feel free to comment on the new commit.

TCFv2/IAB Tech Lab - CMP API v2.md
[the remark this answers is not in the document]
chrispaterson: I like quashing doubt!  Maybe even a little further and to add some reasoning so that people can see the intent.

> Note: The `addEventListener` callback shall be immediately called upon registration with the current TC data, even if the CMP status is `loading` and the CMP has incomplete TC Data, so that the calling script may have access to its registered listenerId.  On every TC String change the callback shall be called unless it is removed via `removeEventListener`.

TCFv2/IAB Tech Lab - CMP API v2.md
chrispaterson: Looks good, I see you found the others as well thanks!

TCFv2/IAB Tech Lab - CMP API v2.md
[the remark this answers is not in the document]
Leedehai: Cool, updated and pushed.

chrispaterson:
Looks good!

wittjill:
@a2intl  also reviewed/approved this

PR #127, opened 2020-01-13
Add TCData validity condition for client side calls

Leedehai:
Suggest how to check TCData for validity. Here validity means the
TC String is considered a true reflection of user's intention of
consent, or lack thereof.


katiestroudpro:
Reviewed and updated.

Leedehai:
Hi @katiestroudpro it looks the implementation guideline doc is updated by adding text directly instead of by merging this PR. So does it mean we can close this PR now?

katiestroudpro:
Yes, we can close.

Leedehai:
Hi, it turns out the update in the PR is not merged. Re-opening this.

PR #143, opened 2020-02-04

wittjill:
tcdata example update


wittjill:
katie fixed on master

PR #144, opened 2020-02-05

DianeJane:
Added version parameter to ping sample.
Fix syntax error by adding closing parenthesis (for grouping) and invoking parenthesis for IIFE.


wittjill:
Thanks @DianeJane - we are actually going to reference the stub in the js library instead as this will be better maintained as an industry resource: https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/pull/145    So I'm going to close this PR out.

PR #158, opened 2020-02-20
Update example strings v2

a2intl:
The example strings start with "B", but the version=2 so they should start with "C". 

Note they still don't parse according to https://iabtcf.com/#/decode; if the reviewer thinks these should be fully-valid examples, please say so, and I'll update with newly-created strings (which seemed to be much longer)


chrispaterson:
Thank AA, I think Jill is working on putting some real examples in from iabtcf.com

a2intl:
@wittjill , if you're working on some real consent string examples, I'll close this PR

wittjill:
I should have those done early next week

PR #160, opened 2020-02-21
oob signal updates

wittjill:
I ended up voting against the truth tables, because, I didn't think that we could concisely fit the full expression into a single table (for disclosed vendor) since in a single transaction, the answer changes depending on whether you store or pass the string. I instead just included a clarifying statement that that is what is happening, since it's a little weird with the rest of the TCF, it seems like something that should be explicitly called out.


alextcone:
Close due to deprecation of Global Scope and OOB in #305

PR #161, opened 2020-02-21
move Global CMP List to this doc after GVL spec

wittjill:


TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
```diff
@@ -1933,3 +1940,85 @@ Here is an annotated example of the GVL’s JSON format:
...
+* A list of CMPs detailing:
+  * A Numeric ID which is incrementally assigned and never re-used - inactive CMPs are marked as deleted.
+  * Their Name.
+  * Whether or not the CMP is a commercial service.
+  * If applicable, the date/time after which CMP is considered inactive.
+
+### Where can I access the Global CMP List?
+The GCL is in JSON format and the current version at any given time can be retrieved using the following URL:
+
+https://cmplist.consensu.org/cmp-list.json
```
dmdabbs: I fetched this and it is the current (v1) CMP list. I thought there was discussion/agreement that there would be a v2-specific URL.

wittjill: updated to include the v2 specific url

patrickverdon:
All looks good to me, thanks.

PR #162, opened 2020-02-21
remove Global CMP list from this doc

wittjill:
moved Global CMP list spec to the tc string and gvl spec doc


patrickverdon:
It might be worth removing the reference to the CMP list in the version history, or otherwise updating the version history to indicate that the CMP list has been moved. Otherwise looks good.

wittjill:
makes sense, included note in change log

PR #165, opened 2020-03-02
Specify callId data type

Leedehai:
This PR clarifies what data type `callId` should have.

To be consistent with `listenerId`, it should be a number, but the code snippet given in the doc implies it is a string. To be backward-compatible, it's reasonable to allow both types.

Note, however, that a calling script should choose one type and stick to it, because JS objects don't differentiate numeric keys from string keys: [MDN doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors) section "property names".


chrispaterson:
This looks great!  I didn't know that IETF RFC-2119 existed, thanks for pointing that out!

Issue #187, opened 2020-04-07
CMP's must now host their own copy of the global vendor list


a2intl:
There is now a requirement that CMP’s serve the vendorlist from their own content server (since the CORS headers aren’t being output on the v2 vendorlist URL, which makes it impossible to fetch via XmlHttpRequest). However, the vendorlist spec, implementation guidelines, and policies don’t mention this (apparently new) requirement. In fact, the vendorlist spec goes into depth about how the cache headers must be respected, which doesn't seem applicable now, at least not from a user's browser point-of-view. The vendorlist spec, CMP implementation guidelines, and policy (if necessary) should be updated with this new requirement. There's also a new requirement (that needs to be specified) of how soon after a new version of the GVL is published that CMP's are required to check for, update their version, and start using it.

alextcone:
Note this is now being discussed on #193 (please add any feedback there)

PR #191, opened 2020-04-14
Clarify CMP hosting and caching

alextcone:
Per issue #187 clarify language for CMPs to understand requirements for hosting, fetching and caching the GVL.

### DIFF
--- a/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
+++ b/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
@@ -1717,6 +1717,8 @@ Given the scale of the TCF and the high volume of requests for the Global Vendor
 
 #### CMPs caching the GVL
 
+CMP code running in browsers should not fetch the GVL directly. The most current version of the GVL must be hosted and provided by each CMP, fetched by the server-side application no more often than the cache-control headers indicate.
+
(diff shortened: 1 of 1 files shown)


a2intl:
A couple issues with this wording and usage of cache control headers (and `max-age` specifically):

The cache-control headers (more specifically, `max-age`, which we should call out explicitly) is defined by RFC2616 as the age at which an item is no longer considered valid (or "stale") and needs to be rechecked. It seems we are repurposing `max-age` to have a different meaning of "the minimum amount of time before rechecking is permitted"; if so, this should be spelled out explicitly. It may be better to instead use a newly-defined [Cache Control Extension](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.9.6) (perhaps `must-cache-for=_numseconds_`), or a new header entirely, perhaps `X-Must-Cache-For`, as it doesn't seem RFC2616 defines a built-in way to specify a minimum amount of time allowed before rechecking.

Also, there's a logical conflict between a requirement to host the "most current version" and fetching it "no more often than the cache-control headers indicate", as it is necessarily a cached version. The freshness-required (i.e. maximum refresh period allowable) needs to be specified, and could be specified to be the `Cache-Control: max-age` value returned (as that follows its defined semantics).

alextcone:
@patrickverdon made his own PR with quite a few more touches on the docs in question. Let's abandon #191 in favor of #193. If you have comments, please put them there.

alextcone:
Closing in favor of #193 which resolves this PR and issue #187

PR #193, opened 2020-04-20
Proposed GVL usage updates

patrickverdon:
I've overhauled and restructured the text around GVL usage to make it clearer.

### DIFF
--- a/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
+++ b/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
@@ -1697,19 +1697,23 @@ As of the publication of this document, changes to the Global Vendor List are pu
 
 ### CMPs using the GVL
 
-Any time the CMP user interface is surfaced to a user to provide transparency and request consent, the **latest** version of the GVL must be used to populate the user interface. This includes first-time interactions, renewal interactions, as well as when a user launches the user interface to manually review and update their settings.
+CMPs must use the **latest** available version of the GVL whenever the interface is surfaced to the user to provide transparency or request consent. This condition applies to first-time and renewal interactions, as well as interactions that involve reviewing and updating settings.
 
-Note that subject to the caching requirements stated below, it is understood that there may be a delay of up to the maximum cache interval in retrieving the latest version of the Global Vendor List, which means that in some cases the version of the GVL used may be the penultimate version of the GVL rather than the latest.
+In some cases, due to caching requirements and low connectivity environments, such as for mobile in-app experiences, it may not be possible to use the latest version of the GVL:
 
-Within a mobile in-app context where the latest version of the GVL cannot be loaded because of a lack of Internet connectivity, the most recently cached version of the GVL may be used – the latest version of the GVL must be retrieved as soon as connectivity is restored.
+* For a delay caused by caching requirements, the penultimate version of the GVL may be considered the latest available version.
+* For a delay caused by a lack of connectivity, the last cached version of the GVL may be used but must be updated as soon as connectivity is restored.
 
-CMPs must use specific versions of the GVL to determine if a CMP should be resurfaced to a user who has a TC String encoded with a GVL version that is not the latest. Note however that it is not a requirement to resurface the CMP if the GVL has changed - when to resurface the CMP user interface is at the discretion of the CMP and publisher.
+To determine whether the interface should be resurfaced to a user, the CMP must compare the latest version of the GVL with the archived version of the GVL identified in the TC String (assuming they are different). The CMP is not required to resurface the interface to the user if the versions are different. The timing and reasons for resurfacing the interface to users is at the discretion of the CMP and the publisher.
 
 **Strict restrictions on accessing and caching the GVL apply and are detailed below.**
 
 ### Vendors using the GVL
 
-Vendors must use the version of the GVL encoded in the TC String received to determine if they have the legal bases they need to process the user's personal data, as well as to determine if a vendor they are about to pass personal data to, also has the necessary legal bases to process personal data.
+Vendors must use the version of the GVL encoded in the TC String received to:
+
+* Determine if they have the legal bases they need to process the user's personal data.
+* Determine if any vendor they are about to pass personal data to, also has the necessary legal bases to process personal data.
 
 **Strict restrictions on accessing and caching the GVL apply and are detailed below.**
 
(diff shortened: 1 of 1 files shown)


katiestroudpro:
I updated some of the language in Patrick's edit, but I wasn't sure if I should edit the file directly. The following text is my suggested edit. Let me know if you'd like to change anything or if you'd like me to incorporate it into the file. Happy to do it. 

Thanks,
Katie


### CMPs using the GVL

CMPs must use the latest available version of the GVL whenever the interface is surfaced to the user to provide transparency or request consent. This condition applies to both first-time and renewal interactions, as well as interactions that involve reviewing and updating settings.

The CMP server-side application pulls the latest available version of the GVL from vendorlist.consensu.org and then passes it to the CMP client-side. With caching requirements and low connectivity environments, such as for mobile in-app experiences, a delay in access may prevent the CMP from providing the latest version of the GVL.

For a delay caused by caching requirements, the last cached version of the GVL may be considered the latest available version and must have been cached within the maximum cache interval.

For a delay caused by a lack of connectivity, the last cached version of the GVL may be used but must be updated as soon as connectivity is restored.

To determine whether the interface should be resurfaced to a user, the CMP must compare GVL versions to track whether a TC String is encoded with the latest GVL version. The CMP is not required to resurface the interface to the user if the versions are different. The timing and reasons for resurfacing the interface to users is at the discretion of the CMP and the publisher.

Strict restrictions on accessing and caching the GVL apply and are detailed in the following sections.


### Vendors using the GVL

Vendors must use the GVL encoded in the TC String received to determine if they have the legal bases they need to process the user's personal data. The same GVL must be reviewed to determine whether any vendor’s downstream have the legal bases to process the user’s personal data.

Strict restrictions on accessing and caching the GVL apply and are detailed in the following sections.


### Accessing and caching the Global Vendor List
In TCF v1 it was possible for client-side CMP applications to load the GVL directly via CORS. Given the scale of the TCF and the high volume of requests for the Global Vendor List, this is no longer possible from TCF v2.0 onward. All requests for the GVL must now be server-side.

Also in TCF v2.0, all current and previous GVL resources, as well as purpose translations, are configured with [cache-control headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control). All requests for these resources must honor the headers and must not cache a resource with different settings, or request a resource more than once within the specified cache period.

Previous versions of the GVL must be cached for at least the period specified by the cache-control headers and may be cached indefinitely since they are static resources.


### CMPs accessing and caching the GVL
Client-side CMP applications must not load the GVL directly from vendorlist.consensu.org - instead the GVL must be loaded and hosted by a CMP’s server-side application and then passed to the client-side CMP application.

CMP server-side applications must only make one request for the GVL during the cache period specified in the cache-control header. For example, if the caching period is one week, only one request for the latest GVL file must be received for a given CMP per week. 

**Important:** The volume of usage will be monitored carefully by the managing organization (MO) and any CMP not adhering to this request limit will be blocked from accessing the GVL.

To prevent client-side applications from repeatedly downloading the file, CMPs should set cache-control headers on HTTP responses sent to client-side requests for their self-hosted GVL resources. When cache-control headers are set, browsers will automatically cache the GVL file. When the client-side application request the GVL file, the browsers return the file from the cache. This process circumvents the need to fetch the file over HTTP multiple times.


### Vendors accessing and caching the GVL

Vendor requests for GVL resources must be loaded and cached by the server-side application.

Vendor application logic must only make one request for the GVL during the cache period specified in the cache-control header. For example, if the caching period is one week, only one request for the latest GVL file must be received for a given vendor per week. 

**Important:** The volume of usage will be monitored carefully by the managing organization (MO) and any vendor not adhering to this request limit will be blocked from accessing the GVL.

jawadst:
@patrickverdon @katiestroudpro I understand the IAB Europe not wanting to host the GVL for all websites/mobile apps out there but we should keep requirements more open to various implementations. The IAB should dictate the end result and not how it gets achieved.

1) On the version being current, the TCF validator indicates that a CMP is allowed to use the latest 2 versions (ie the latest or the previous). Allowing CMPs to have some delay to update (latest 2 versions being more or less a 1-week delay) has practically no impact on the TCF (the GVL updates are not important enough that they can't wait a day or two) and allows CMPs to implement that the way they think is the most efficient for their publishers and end-users.

The requirement "CMPs must use the latest available version of the GVL whenever the interface is surfaced to the user to provide transparency or request consent." does not seem realistic to me. CMPs are trying to optimize their loading times by doing a lot of caching, static serving, etc. which is more important from a user experience perspective than being on the absolute latest version at all times.

2) CMP server-side applications must only make one request for the GVL during the cache period specified in the cache-control header. 
I understand that this requirement is about not passing on the volume from the CMP to the IAB Europe but an "exactly one" request for the cache period is really impractical and uncommon. That requirement should be expressed in a way that protects the IAB Europe but allows CMPs to implement an efficient hosting/caching solution. For instance, no more than 1 request per second from a given CMP/publisher. "Exactly one" is always hard to guarantee in a distributed environment and 1 request per minute or 50 requests per minute has virtually no cost impact on the IAB Europe. I am sure we can find a better balance here.

alextcone:
@patrickverdon do you have thoughts on @jawadst's feedback? (also, thanks @katiestroudpro for jumping on this)

TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
```diff
@@ -1697,43 +1697,43 @@ As of the publication of this document, changes to the Global Vendor List are pu
...
 
-#### Vendors caching the GVL
+### CMPs accessing and caching the GVL
 
-Because vendor requests for a GVL file will not be in a browser context, GVL files must be cached server-side.
+Client-side CMP applications must not load the GVL directly from vendorlist.consensu.org - instead the GVL must be loaded and hosted by a CMP’s server-side application and then passed to the client-side CMP application.
 
-Vendor application logic must only request one version of the GVL per vendor during the cache period specified in the cache-control header. For example, if the caching period is one week, only one request for the current GVL file must be received for a given vendor per week.
+CMP server-side applications must only make one request for the GVL during the cache period specified in the cache-control header. For example, if the caching period is one week, only one request for the latest GVL file must be received for a given CMP per week. Note: The volume of usage will be monitored carefully by the managing organisation (MO) and any CMP not adhering to this request limit will be blocked from accessing the GVL.
```
a2intl, line 1728: There are two time periods that need to be communicated (and enforced by this spec): the minimum amount of time that is allowable between fetches, and the maximum amount of time that a fetched GVL can be considered "fresh". We cannot communicate both these time periods with a single value in `max-age`, as they need to be different values-- the minimum time allowable between fetches is necessarily less than the amount of time the GVL can be considered fresh.

a2intl: (for instance, the minimum time allowed between fetches should be on the order of a day, but the maximum time that the GVL is considered fresh should be it's expected update interval, i.e. a week)

a2intl: Otherwise, we are enforcing the impossible-to-meet scenario of "you can only check once a week, but your GVL must not be older than a week".

patrickverdon:
@jawadst 

> 1. On the version being current, the TCF validator indicates that a CMP is allowed to use the latest 2 versions (ie the latest or the previous). Allowing CMPs to have some delay to update (latest 2 versions being more or less a 1-week delay) has practically no impact on the TCF (the GVL updates are not important enough that they can't wait a day or two) and allows CMPs to implement that the way they think is the most efficient for their publishers and end-users.

The Validator allows for the latest 2 versions due to the cache-control headers being set to one week. For example, if the GVL is requested 5 minutes before the new version is made available and cached for 1 week as required, the penultimate version of the GVL will always be used.

> The requirement "CMPs must use the latest available version of the GVL whenever the interface is surfaced to the user to provide transparency or request consent." does not seem realistic to me. CMPs are trying to optimize their loading times by doing a lot of caching, static serving, etc. which is more important from a user experience perspective than being on the absolute latest version at all times.

The requirement to show the latest GVL is a policy requirement, however it is subject to the caching requirements, so it is acceptable to use the penultimate GVL version.

> 1. CMP server-side applications must only make one request for the GVL during the cache period specified in the cache-control header.
>    I understand that this requirement is about not passing on the volume from the CMP to the IAB Europe but an "exactly one" request for the cache period is really impractical and uncommon. That requirement should be expressed in a way that protects the IAB Europe but allows CMPs to implement an efficient hosting/caching solution. For instance, no more than 1 request per second from a given CMP/publisher. "Exactly one" is always hard to guarantee in a distributed environment and 1 request per minute or 50 requests per minute has virtually no cost impact on the IAB Europe. I am sure we can find a better balance here.

This is what has been agreed and to be honest, I can't see a good reason for server-side caching to make more than 1 request per the caching period. The server-side application simply needs to behave like a browser. This is pretty straightforward using memcached, Redis, Varnish or any other caching infrastructure.

TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
patrickverdon: What we're trying to convey is that the server-side application needs to behave like a browser - so the back-end needs to do something along the lines of:

1. Set-up a cronjob that requests the GVL
2. Cache the resource for the period specified by max-age
3. At max-age+1 clear the cache and re-request the GVL

This can be easily done with a proxy like Squid or Varnish, or manually using memcached, etc.  

So the only time period we need to communicate is ```max-age```, as would be communicated to a browser. The minimum amount of time allowable between fetches is the ```max-age``` value. And the "freshness" of the GVL will be up to 2 weeks in line with the current caching policy of 1 week, which will mean the latest or penultimate version of the GVL is used.

I've updated the text to make this clearer.

PR #206, opened 2020-05-27
Adding union type of false | undefined as undefined means no consent to TCData and InAppTCData

chrispaterson:
To avoid requiring the GVL to be downloaded when decoding the TCString inside the CMP API before any TCData request `false` values are equivalent to `undefined` values.


HeinzBaumann:
Thanks @chrispaterson for updating the specification regarding this.

PR #212, opened 2020-06-16
Handling conflict between v1.1 strings and v2 strings

alextcone:
During a Framework Signal WG call deprecation of TCF v1.1 strings was discussed and it was agreed that help manage this process and maintain the integrity of the TC String (the WG mission) that CMPs should be asked to remove TCF v1.1 strings if a TCF v2.0 string was present. This ensures that vendors are not conflicted by presentation of two strings. Policy have have confirmed that this does not require a policy change but we feel that this does require a Tech Spec change. Patrick is off this week so I am picking this up in his absence.

### DIFF
--- a/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
+++ b/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
@@ -387,7 +387,7 @@ The following table summarises where data is stored:
 </table>
 
 #### Managing conflicting string versions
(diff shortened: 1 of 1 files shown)


alextcone:
@ grantimus9

alextcone:
Address Will's comment that v1.1 is a bit specific, perhaps add v1 reference as those could erroneously exist in the wild as well...as long as we're being clear.

alextcone:
Be clear that CMPs shouldn't nuke a global cookie before the transition period is over.

PR #220, opened 2020-07-27
Update publisher restrictions format

janwinkler:
added "purpose 63" (to be discussed) in order to signal that this override should apply to all purposes. an "override all" is the more common scenario but is currently requiereing 10 restriction sections for each of the 10 purposes. an "all purposes" type can reduce the length of the tc string drastically while still being very explicit in the meaning.

### DIFF
--- a/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
+++ b/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
@@ -1022,7 +1022,7 @@ CLcVDxRMWfGmWAVAHCENAXCkAKDAADnAABRgA5mdfCKZuYJez-NQm0TBMYA4oCAAGQYIAAAAAAEAIAEg
       <td>Purpose ID</td>
       <td>
         The Vendor’s declared Purpose ID that the publisher has indicated
-        that they are overriding.
+        that they are overriding. Use Purpose ID 63 in order to signal "all purposes".
       </td>
     </tr>
     <tr>


janwinkler:
maybe purpose=0 makes more sense

patrickverdon:
Having an 'all purposes' option may be confusing where purposes have either a) not been registered by a vendor or b) have not been registered as flexible, however if this is a common use case then it would definitely help to reduce the size of the TC String.

chrispaterson:
I prefer purpose 0 to 63 (because it's more stable and easier to remember).

To @patrickverdon 's point for restrictionType 1 and 2, it would only apply to purposes declared as flexible by the Vendor.  That should probably be stated in the text.

janwinkler:
yes, after rethinking it several times, purpose=0 makes more sense.

janwinkler:
Quick background info: We had a publisher adding so many restrictions that the TC string was 12 KB large. if we only use purpose=1, the  tc string would have been around 500byte ...

matita:
I agree with @chrispaterson about purpose `0` being better for a number of reasons.

dmdabbs:
As I understand from above, the digital property wants to restrict _all purposes_ for some set of vendors. I don't understand why we need a specification change. The digital property should direct their CMP to _not disclose these vendors_. Vendors that are not disclosed should have no signals in the string and so can't process based on it.

janwinkler:
@dmdabbs the use case is not to remove vendors but to restrict all vendors and all purposes to consent. in this case we have 10 parts (one for each purpose) that lists all vendors and sets them to consent. in order to save bytes we set purposeid=0 and then list all vendors. this should express the same "all vendors should use consent (only) for all purposes if vendor declared that purpose to be flexible" (or we can even go further and say "all vendors should use consent (only) for all purposes if vendor declared that purpose to be flexible AND not use the purpose at all if the vendor did not declare it as flexible")

janwinkler:
actually the "all vendors and all purposes" could be even more packed if we allow the ranges to also contain id's that do not exist: e.g. currently the range will look something like: "1-3,5,27-29,31" in order to cover vendors 1,2,3,5,27,28,29,31. but we could reduce the size if we allow a range to be only "1-31". but lets talk about the purposeid=0 first and get it step by step

dmdabbs:
>the use case is not to remove vendors but to restrict all vendors and all purposes to consent.
I see. Not all vendors use 'flexible' declarations. What of them?

janwinkler:
@dmdabbs what we do with the non-flexible vendors is up for debate. either we understand it as "they are not allowed for this purpose" or "they are not meant with this". its basically the same question: what is the current interpretation if a tcstring contains "restrict purpose 3 on vendor 4 to consent" if vendor 4 did not declare the purpose as flexible (or did not decalre it at all)?

dmdabbs:
>what is the current interpretation if a tcstring contains "restrict purpose 3 on vendor 4 to consent" if vendor 4 did not declare the purpose as flexible (or did not declare it at all)?

There probably insn't anything explicit since this signal feature was intended to be used in a precise manner. Since the Policies - and this feature - do not provide a means for a property to force a registered Vendor to process in a way that they haven't registered their willingness to in their GVL registration, signaling in that case should be ignored IMO. If the vendor didn't declare flexibility for that purpose then this signal does not apply to that vendor. String consumers should take care to properly check against the GVL.  We should add appropriate spec guidance if/when we describe the proposed "Purpose 0" scheme. Same if the vendor didn't register for the purpose at all in the first place. Especially so if there isn't an active vendor with a covered id (due to GVL numbering gaps).

dmdabbs:
If it's not already covered, then I think that we should review how the user-visible disclosures should be handled in the presence of restrictions. I think the user must see what the property is restricting, since that those are the signals vendors and intermediaries will be expected to inspect. In your case, if the site is limiting/coercing the flexible vendors to use consent, then the CMP must present consent UX for those vendors' purpose disclosures so restricted.

janwinkler:
@dmdabbs why is it important to the visitor to know that the site restricted the vendors? as long as the right mechanism (e.g. consent toggle vs LI opt out toggle) is presented, the visitor shouldnt really care about the WHY, or do I miss something?

asr-enid:
Restrictions are applied prior to showing the UI, users need to be aware of the specific legal ground for a purpose and/or if a purpose is allowed at all (he will not care if its restricted or simply not registered by a vendor)

dmdabbs:
I believe we agree. When I wrote "the user must see that the site is restricting" the phrase I should have used is "the disclosures and user choices should be aligned to the restrictions," which is what I then went on to describe.

matita:
@dmdabbs 
> > what is the current interpretation if a tcstring contains "restrict purpose 3 on vendor 4 to consent" if vendor 4 did not declare the purpose as flexible (or did not declare it at all)?
> 
> There probably insn't anything explicit since this signal feature was intended to be used in a precise manner.

Actually in the TC string specs there is an explicit clarification about how vendors that did not specify flexibility should deal with the restriction signal https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20Consent%20string%20and%20vendor%20list%20formats%20v2.md#the-core-string (Publisher Restrictions Section/RestrictionType).

dmdabbs:
I don't believe we require any specification changes to "signal to all vendors that they may not process under legal basis X" using the smallest possible strings, To be clear, this applies only to processing for Purposes and not Special Purposes, since the latter are always LI and I would think Special Purpose processing is non-controversial. 

We've agreed that the CMP UX will be commensurate with the property's (publisher) restrictions (applied when showing the UI). 

If the property instructs its CMP to "signal to all vendors that they may not process under legitimate interest," then the following should happen:
- Should be no need to display the required LI wording e.g. "some processing based on legitimate interest but you have a right to opt out" (since no LI processing will be disclosed and the signals will indicate that it is not permitted).
- No legitimate interest UX is provided at the purpose level, only consent.
- No legitimate interest UX is provided at the vendor level, only consent.
-- If a Vendor has registered a Purpose as flexible, then the Purpose will be disclosed in that Vendor's purposes section as using consent.
-- If a Vendor has registered a Purpose as only legitimate interest, then the Purpose will not be disclosed at all in that Vendor's purposes section.

Given this UX, the tcstring the CMP produces _will have no signals that permit use of legitimate interest_.
- PurposesLITransparency will encode all 0s.
-- This alone should be sufficient to block all vendors' legitimate interest processing. 
- The "Vendor Legitimate Interest Section" should be empty. Since the section is not optional, this is a mechanical exercise in finding the most compact encoding that is effectively 'empty.' For example: MaxVendorId=0, IsRange=1, NumEntries=0

So no need for any "Publisher Restrictions Section" spec changes or use of "Purpose 0" for _this specific use case_. 

This should be in accord with Policy 23.3 (others may apply, but this seems most germain):
>23.3. A Publisher may choose which Purposes, SpecialFeatures, and/or Vendors to disclose. If a Publisher chooses not to disclose a Purpose, SpecialFeature, and/or a Vendor, the Signals must appropriately reflect in the Signal that no LegalBases and/or opt-ins have been established for the respective Purposes, SpecialFeatures, and/or Vendors. For the avoidance of doubt: SpecialPurposes, and Features must always disclosed if at least one of the Vendors disclosed has declared to be using them.
>23.4.A Publisher may restrict certain Purposes for specific Vendors, these restrictions must be implemented by the CMP, which shall reflect Publisher restrictions in both the User Interface and the Signals in accordance with the Policies and Specifications.

williamleeMM:
I agree with David's reasoning and conclusion in the above comment that a spec change is not needed here. A spec change this late in the game is also less than ideal.

ptrickney:
I agree with David's reasoning as well, however, we've heard that vendors are not necessarily interpreting it this way. They are interpreting that the only way for them to use their flexible purpose is if there is a publisher restriction. 

For example, if the vendor has Purpose 2 as a flexible purpose with Leg. Int. as default, then if they do not get the Leg. Int. signal they will not fall back to the consent legal basis unless they see a publisher restriction to require consent.

williamleeMM:
Thanks @patrickverdon. I must have been updating my previous comment just as you responded.

Agree that's a concern, but I think it's best solved by clarification in the specification commentary and outreach.

dmdabbs:
I described an approach for businesses who feel it's necessary to "completely turn off legitimate interest" using available signaling in a space-efficient manner _business impact be damned_ (since flexible vendors will not get the restriction signal to check their alternate basis). 

We'll get the Commit Group or designated sub-team to quarterback a spec change solution. 

In the mean time, is this an accurate inventory of the choices?

- Go with legitimate interest while we cook up an official, space-efficient approach to this use case.
- Turn off legitimate interest across the board 
- Turn off legitimate interest across the board (but identify key vendors who will receive the restriction signal so your strings aren't too large)

dmdabbs:
Digital properties (publishers) are telling you they want to prohibit all use of legitimate interest for all Purposes. Special Purposes are no problem. But encoding all this precisely has resulted in prohibitively large tcstrings. So you want a space-efficient encoding.
The following should meet the need and requires no spec encoding change. An update to the spec to discuss this specific scenario is advised.

#1
Set all PurposesLI field bits to 0.
This prevents any Vendor from using legitimate interest for any Purpose.

#2
Encode an effectively empty "Vendor Legitimate Interest Section."
MaxVendorId 65536
IsRangeEncoding 1
NumEntries 1
IsARange 0 (a single entry)
StartOrOnlyVendorId 65536

NOTES:
There's no Vendor id 65536 and that's the point. This is a harmless bit of inexactitude.
If we decide to formally make this section optional, we should explicitly permit MaxVendorId = 0 to signal "nothing else encoded here, proceed to decoding the next section."

#3
Encode 9 "Publisher Restrictions Section" entries as follows:
NumPubRestrictions 9
PurposeId 2 
RestrictionType 1 (Require Consent (if Vendor has declared the Purpose IDs legal basis as Legitimate Interest and flexible))
NumEntries 1
IsARange 1
StartOrOnlyVendorId 1
EndVendorId 811 (811 is the highest Vendor id in the latest GVL)
(repeat for Purposes 3 through 10)

NOTES:
More inexactitude in service to encoding efficiency.
For Vendors that
* have not registered this purpose at all the entry is meaningless.
* registered this purpose for consent legal basis the entry is meaningless.
* registered this purpose for legitimate interest, but not flexibly, the entry is meaningless.
* registered this purpose for legitimate interest but are flexible, they will know to read consent signals for this purpose as described in the specification. This is the original intended use case for this signal.

The 1 to max vendor range will include gaps in vendor id numbering. Should IAB assign one of these gap id to a new Vendor during the life of this string and that vendor happens to register flexibly for any Purpose, then this string will ensure that the Vendor may not use legitimate interest just as the Publisher intended.
No need to encode an entry for Purpose 1 since IAB does not permit Vendors to register it using legitimate interest.

It should be possible to add a convenience method to the standard library to accommodate this scheme.

chrispaterson:
If the publisher would like to disallow all legitimate interest on their site.  All legitimate interest vectors – vendor and purpose – will be set to negative signals.  To reduce size of the publisher restrictions TC String segment, encode the vendor ranges within the purpose/type restriction to include vendors who have no flexibility; the spec explicitly directs vendors to ignore a type 1 or 2 restriction if they are not flexible.

Purpose 0 would save a total of 64b¹ in total in the above scenario, so it could be added as a further optimization but given the risk and effort to message the change it may not be worth amending the 2.0 spec.

1. **A Purpose Restriction segment is 7 bytes, 9 purposes may be overridden**:
7b  x 9 = 63b * (4/3 for base64 encoding) = **84b**
**Purpose Restriction segment is 7 bytes, 1 "purpose 0"**:
7b x 1 = 7b * (4/3 for base64 encoding) = **10b**
**Difference**
84b - 10b = 64b

alextcone:
@dmdabbs - re: "An update to the spec to discuss this specific scenario is advised." <= where would you suggest this live?

chrispaterson:
> @dmdabbs - re: "An update to the spec to discuss this specific scenario is advised." <= where would you suggest this live?

My vote would be the implementation guidelines because the information is all there but it's just maybe not all pulled together for some.

asr-enid:
> My vote would be the implementation guidelines because the information is all there but it's just maybe not all pulled together for some.

+1 - it would be extremely valuable to have a more extensive guidance on:

- usage of publisher restrictions and how they relate to the UI portion/CMP config (this has been answered many times, but it is coming up over and over again)
- interpretation if restricted signals by vendors

@chrispaterson would be great to have hands on decode / encode examples, for example 

1. loading the GVL
2. narrow the vendors
3. apply restrictions (flexible vendor, non-flexible vendor, disallow)
4. display UI and set choices via the api
5. encoded result for storage, as well as signal sent

maryxiaoyong:
+1 to David's proposal of encoding above and Chris's assessment on minimal size savings re: purpose0.

chrispaterson:
> Given it checks `if (this.isOkToHave(purposeRestriction.restrictionType, purposeRestriction.purposeId, vendorId))` it won't add those with no flexibility and just create a large number of segments instead one range (given its a scattered list)

I think this would end up being an encoding detail that would look for the longest span of unaffected vendors (ie. not flexible or non-existent).  In decode, it would simply reverse the process.

alextcone:
Where do we stand on this?

janwinkler:
> Encode an effectively empty "Vendor Legitimate Interest Section."
> MaxVendorId 65536
> IsRangeEncoding 1
> NumEntries 1
> IsARange 0 (a single entry)
> StartOrOnlyVendorId 65536

Did we agree that this is valid? I mean not explicitly telling the IDs but just saying "everything from 1 to 65k"?
Also, shouldnt it rather be the following to signal "no vendor has LI":
MaxVendorID=0
IsRangeEncoding=1
NumEntries=0

janwinkler:
ah sorry, misread the example. anyhow, endcoding NumEntries=0 would be better than sending NumEntries=1 and a fake vendor ID

janwinkler:
my opinion on the space saving:
1. allow purpose=0 to signal "all purposes"
2. allow range 0-65k in order to signal "all vendors" (instead of naming each ID)

janwinkler:
> Purpose 0 would save a total of 74b¹ in total in the above scenario, so it could be added as a further optimization but given the risk and effort to message the change it may not be worth amending the 2.0 spec.

only if a range [1 - 899] is allowed, which from my point of view is currently not allowed by the spec ...

Facens:
I put together a proposal that would allow us to shring the string by 77% in case of restrictions without making any changes to the spec: https://docs.google.com/document/d/1yoYN-iD7wlLjPt1DyEYN8KFAyiO_yZGk_CWhiWOnqAI/edit#

I'm looking for consensus around this proposal as it would be readily available for everyone to benefit from and would require no changes for vendors nor for CMPs (except CMPs that want to optimize the size).

At the same time, we can discuss more structural optimizations, like allowing ranges without having to account for gaps, if there are no drawbacks for it.

Let me know what you think.

asr-enid:
@Facens - I might be mistaken, but if the proposal is to only add a vendor to a restriction in case it is actually needed (as described) - that is something that @chrispaterson has already implemented inside the encoder/TCModel. If that is applied for all purposes (which is the case we are talking about here), this still leads to too large TC Strings. Just Encoding 1-Max as Chris suggested would be way smaller (but is no possible with the current implementation)

When a CMP adds a restriction the library checks if its necessary 

https://github.com/InteractiveAdvertisingBureau/iabtcf-es/blob/07cd5c1e0d2f3ac6fb86f3ad1010d4721f8ca822/modules/core/src/model/PurposeRestrictionVector.ts#L108

`if (this.isOkToHave(purposeRestriction.restrictionType, purposeRestriction.purposeId, vendorId)) {...`

_isOKToHave_ checks what you are describing:

https://github.com/InteractiveAdvertisingBureau/iabtcf-es/blob/07cd5c1e0d2f3ac6fb86f3ad1010d4721f8ca822/modules/core/src/model/PurposeRestrictionVector.ts#L31

Case 1:

` /**
             * If the vendor has the purposeId in flexiblePurposes and it is
             * listed as a legitimate interest purpose we can set the
             * override to require consent.
  */`

Case 2:

` /**
             * If the vendor has the purposeId in flexiblePurposes and it is
             * listed as a consent purpose we can set the
             * override to require legitimate interest.
*/`

Facens:
We just checked and this was actually recently changed in the SDK, @chrispaterson do you confirm? @alextcone I wonder if it's worth that we also specify it in the guidelines.

alextcone:
@facens, I’m always pro clarifying guidelines. Will you make a PR on the doc?

asr-enid:
Just to avoid redundancy - I think @chrispaterson is looking into how his thoughts could be formalised which go beyond what is already implemented. https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/pull/220#issuecomment-675056333

Unrelated to that it would be very helpful to explicitly mention in the docs when the overrides must be in place in case the publisher wants to change the applied legal basis and that in case there is no override that the default legal basis always applies.

asr-enid:
See PR for an extended description on restrictions in the tc string format document. Its basically a less details/non technical version of what is in the format table description and I hope that it covers the points discussed here.

asr-enid:
@alextcone @chrispaterson - how do we progress with the encoding question? With the merged guidance it is clear now when restrictions are needed, the issue with too large TC Strings in case publishers want to restrict a whole purpose for all vendors (if needed) to one legal basis still remains.

alextcone:
@asr-enid We can bring it to the Global Privacy Working Group. Do you think we can address it in a way that doesn’t create a breaking change? I’m thinking about the current path of moving towards a more generic backend for TCF and how that path may overlap with encoding improvements and the need for them. Do you see these two areas as potentially interrelated?

asr-enid:
@alextcone - My understanding is that there is a larger consensus that the current spec would support an efficient way of encoding this into the string without breaking changes for vendors. I'd suggest to

1. Re-Align that this is the case (@dmdabbs @chrispaterson already expressed that above), fix the final suggested version (Range Entry Setup)
2. Add this as a convenience method to the libraries, not only for convenience but because the current version would not allow those entries to be written into the string (as it removes non strictly necessary restrictions for good reasons)
3. Extend the vendor/cmp guidance to explicitly cover this and fix any contradictory guidance if present

I think it would also be worthwhile to include a more explicit declaration of this, but this will always be a breaking change as it ads new semantics that would need to be adopted by vendors. Could be done in case there is a breaking change coming for other means anyhow. 

> I’m thinking about the current path of moving towards a more generic backend for TCF and how that path may overlap with encoding improvements and the need for them. Do you see these two areas as potentially interrelated?

They might be, my personal assumption is that the configurations will get simpler over the course of time in that regard which means he backends would become simpler (i.e. not needing the full matrix of purposes, vendors and legal basis). Anyhow we should define a hands-on solution now, because there will be more and more pubs that will needs this type of configuration.

asr-enid:
I think we can close this one as discussed on the Global Privacy WG Call 11/12/20:

PR for Action Item 1 - https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/pull/253

alextcone:
Closing per @asr-enid and the fact that we're going to handle the code change in https://github.com/InteractiveAdvertisingBureau/iabtcf-es 

> I think we can close this one as discussed on the Global Privacy WG Call 11/12/20:
> 
> PR for Action Item 1 - #253

PR #225, opened 2020-08-06
Deprecate cmpStatus = loading

alextcone:
This status is not distinct or descriptive and is therefore considered deprecated. It will be removed in the next version.


chrispaterson:
Looks good to me!

PR #230, opened 2020-08-14
Clarify GVL example language

alextcone:
Fix typo, re-arrange and add a few lines for clarity


alextcone:
Originally raised by Jacob Grabczewski (Xaxis) on August 14, 2020

PR #244, opened 2020-09-19
Clarification on restrictions

asr-enid:
As discussed in https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/pull/220 - some more explicit description in the section "What are publisher restrictions?". Looking at the discussion it seem good to make it more explicit 

- when which restrictions must be applied
- to explicitly state that the default must always be applied in the absence of a signal
- legal basis signal without vendor flexibility are redundant.

@chrispaterson the TC Core String details description contains some language in the notes that restrictions **should** be respected even without flexibility, not sure I you'd want to change that given the code would not even allow them to be written just now


asr-enid:
Adding here - the language around default legal basis is already present in the spec, but just in the annotated example global vendor list - seem even more reasonable to give this a bit more visibility. What missing here is that the default can only be changes via a restriction.

`
   * "legIntPurposes" conditionally OPTIONAL (see "Constraints") array of
   * positive integers. List of Purpose ids declared as performed on the legal
   * basis of legitimate interest
   *
   * "flexiblePurposes": OPTIONAL array of positive integers. Array may be
   * empty. List of purpose ids where the vendor is flexible regarding the
   * legal basis; they will perform the processing based on consent or a
   * legitimate interest. **The 'default'** is determined by which of the other two
   * mutually-exclusive purpose fields is used to declare the purpose for the
   * vendor
   *
 * Constraints:
   *   Either purposes OR legIntPurposes can be missing/empty, but not both.
   *
   *   A Purpose id must not be present in both purposes and legIntPurposes
   *
   *   A Purpose id listed in flexiblePurposes must have been declared in one
   *   of purposes or legIntPurposes.
   *
   *   Purpose id values included in the three purpose fields must be in the
   *   range from 1 to N, where N is the highest purpose id published in this
   *   GVL file.`

https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20Consent%20string%20and%20vendor%20list%20formats%20v2.md#example-global-vendor-list-json-object

Facens:
@asr-enid thanks, the changes look good to me. What about also adding to the CMP guidelines that CMPs can avoid sending redundant information, as per the gdoc here? https://docs.google.com/document/d/1yoYN-iD7wlLjPt1DyEYN8KFAyiO_yZGk_CWhiWOnqAI/edit?disco=AAAAG15TmLU&ts=5f720228&usp_dm=false
I can review if you take the lead and add it, so we can get this moving forward quickly. Thanks!

asr-enid:
@Facens - you mean within the general implementation guidelines? https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/TCF-Implementation-Guidelines.md#cmp

Facens:
@asr-enid Yes! This is what we've agreed on the latest group meeting.

asr-enid:
Ok - will add something to this pull request (makes sense given its the same context), would like to get this closed

asr-enid:
Added some language in the implementation guidelines to reflect in which cases restrictions are necessary. Added some language that this does not preclude encoding schemes that might deviate (as suggested by @chrispaterson ) from this because they might be more space efficient with the same results

asr-enid:
@Facens - let me know what you think - I'd find it highly valuable to have both clarifications merged soon.

TCFv2/TCF-Implementation-Guidelines.md
```diff
@@ -251,7 +252,19 @@ You’ll usually want to go with a combination of server-side storage – for be
 ## 4. Withdrawal of consent and other non-TCF policy <a name="withdrawconsent"></a>
 Signals sent through the IAB Europe framework should only indicate what the user status is at the time of the signal creation and nothing else. While the CMP should also enable users to withdraw consent, the minimum requirement is to record the user's preference at the time the signal is created. Certain GDPR policy, such as portability and the right to be forgotten, is not covered in the IAB Europe TCF. CMPs and vendors should address other GDPR rights outside the TCF separately and on their own.
 
-## 5. CMP interface requirements<a name="cmpreq"></a>
+## 5. Encoding publisher restrictions <a name="pubrestrenc"></a>
+CMPs are advised to store/provide publisher restrictions only in cases they are necessary to reflect the publishers choice to restrict a vendors processing of personal data. In terms of reflecting a publisher’s choice:
```
Facens, line 256:
```suggestion
In order to reduce the size of the TC string, CMPs are advised to store/provide publisher restrictions only when necessary to reflect the publisher's choice to restrict a vendor's processing of personal data. In terms of reflecting a publisher’s choice:
```

TCFv2/TCF-Implementation-Guidelines.md
```diff
@@ -251,7 +252,19 @@ You’ll usually want to go with a combination of server-side storage – for be
...
+
+* In case a vendor has not been disclosed to the user via the CMP UI, there is no need to store restrictions for that vendor in the TC String. Given the vendor was not disclosed both vendor consent and vendor legitimate interest signals in the TC String must be set to 0 which suffices to signal that the vendor may not process personal data.
```
Facens, line 258: must be set to 0 -> can be left undefined?

TCFv2/TCF-Implementation-Guidelines.md
```diff
@@ -251,7 +252,19 @@ You’ll usually want to go with a combination of server-side storage – for be
...
+* Purpose restrictions (i.e. disallowing a vendor to process) need to be stored in case the vendor was disclosed by the CMP (reflecting the restriction in the UI) and registered for that purpose in the GVL
```
Facens, line 259: Not sure I follow this one. I think you're referring to restricting a purpose for a specific vendor, in which case the restrictions{} must be passed with a disallowed. If so it should be clarified.

TCFv2/TCF-Implementation-Guidelines.md
```diff
@@ -251,7 +252,19 @@ You’ll usually want to go with a combination of server-side storage – for be
...
+* Legal Basis restrictions are only needed in situations where there vendor was disclosed by the CMP (reflecting the restriction in the UI)  and declared flexibilty in the GVL for the respective purpose, meaning:
```
Facens, line 260:
```suggestion
* Legal Basis restrictions are only needed in situations where there vendor was disclosed by the CMP (reflecting the restriction in the UI) and declared flexibilty in the GVL for the respective purpose, meaning:
```

TCFv2/TCF-Implementation-Guidelines.md
```diff
@@ -251,7 +252,19 @@ You’ll usually want to go with a combination of server-side storage – for be
...
+
+    1. The vendor registered a purpose as legitimated interest (default legal basis), but also registered this purpose as flexible (i.e. also accepts consent as a legal basis). In this case a "require consent" restriction is needed to signal that the vendor may only apply the consent signal/can rely on it to process personal data.
```
Facens, line 262:
```suggestion
    1. The vendor registered a purpose as legitimate interest (default legal basis), but also registered this purpose as flexible (i.e. also accepts consent as a legal basis). In this case a "require consent" restriction is needed to signal that the vendor may only apply the consent signal/can rely on it to process personal data.
```

TCFv2/TCF-Implementation-Guidelines.md
```diff
@@ -251,7 +252,19 @@ You’ll usually want to go with a combination of server-side storage – for be
...
+    2. The vendor registered a purpose as consent (default legal basis), but also registered this purpose as flexible (i.e. also accepts legitimate interest as a legal basis).  In this case a "require legitimate interest" restriction is needed to signal that the vendor may only apply the legitimate interest signal/can rely on it to process personal data.
+    
+Note that the above does not preclude the use of efficient encoding/decoding schemes in certain scenarios, for example in cases where a publisher wants to restrict a purpose for all vendors to consent it might be more efficient to encode a small number of range restriction segments using an specific encoding scheme.
```
Facens, line 265:
```suggestion
Note that the above does not preclude the use of efficient encoding/decoding schemes in certain scenarios, for example in cases where a publisher wants to restrict a purpose for all vendors to consent it might be more efficient to encode a small number of range restriction segments using a specific encoding scheme.
```

Facens (changes requested):

TCFv2/TCF-Implementation-Guidelines.md
alextcone: It looks like there is alignment here. I'd like to merge this in by the end of the work week. Do you think we're there?

TCFv2/TCF-Implementation-Guidelines.md
asr-enid: Fine for me, adapted accordingly

TCFv2/TCF-Implementation-Guidelines.md
asr-enid: This is the same language as used in the TC String Format descriptions, a restrictions is either 

1. **Purpose** restrictions -> a vendor is disallowed to process for a purpose 
2. **Legal Basis** restrictions -> a vendor is forced to use a different legal basis and may not rely on its default one

Changed the language to make this clearer

asr-enid:
@alextcone I'm good - accepted the suggested changed from @Facens and changed the language on purpose restrictions as well as leaving vendor bits undefined rather them setting them to 0 (last commit).

TCFv2/TCF-Implementation-Guidelines.md
```diff
@@ -251,7 +252,19 @@ You’ll usually want to go with a combination of server-side storage – for be
...
 
-## 5. CMP interface requirements<a name="cmpreq"></a>
+## 5. Encoding publisher restrictions <a name="pubrestrenc"></a>
+In order to reduce the size of the TC string, CMPs are advised to store/provide publisher restrictions only when necessary to reflect the publisher's choice to restrict a vendor's processing of personal data. In terms of reflecting a publisher’s choice:
+
+* In case a vendor has not been disclosed to the user via the CMP UI, there is no need to store restrictions for that vendor in the TC String. Given the vendor was not disclosed both vendor consent and vendor legitimate interest signals in the TC String can be left undefined which suffices to signal that the vendor may not process personal data.
```
Facens, line 258:
```suggestion
* In case a vendor has not been disclosed to the user via the CMP UI, there is no need to store restrictions for that vendor in the TC String. If the vendor was not disclosed, both vendor consent and vendor legitimate interest signals in the TC String can be left undefined which suffices to signal that the vendor may not process personal data.
```

TCFv2/TCF-Implementation-Guidelines.md
```diff
@@ -251,7 +252,19 @@ You’ll usually want to go with a combination of server-side storage – for be
...
+* Purpose restrictions that disallow a vendor to process personal data for a specific purpose only need to be stored in case the vendor was disclosed by the CMP (reflecting the restriction in the UI) and registered for that purpose in the GVL
+* Legal Basis restrictions are only needed in situations where there vendor was disclosed by the CMP (reflecting the restriction in the UI) and declared flexibilty in the GVL for the respective purpose, meaning:
```
Facens, line 260:
```suggestion
* Legal Basis restrictions are only needed in situations where the vendor was disclosed by the CMP (reflecting the restriction in the UI) and had declared flexibility in the GVL for the corresponding purpose, meaning:
```

TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
```diff
@@ -190,12 +190,20 @@ If the disclosures do not describe a global scope, or explicitly state service-s
...
+2. Vendors that declared a purpose with a default legal basis (consent or legitimate interest respectively) but also declared this purpose as flexible must respect a legal basis restriction if present. That means for example in case they declared a purpose as legitimate interest but also declared that purpose as flexible and there is a legal basis restriction to require consent, they must then check for the consent signal and must not apply the legitimate interest signal.
+
+For the avoidance of doubt:
+
+* In case a vendor has declared flexibility for a purpose and there is no legal basis restriction signal it must always apply the default legal basis under which the purpose was registered aside from being registered as flexible. That means if a vendor declared a purpose as legitimate interest and also declared that purpose as flexible it may not apply a "consent" signal without a legal basis restriction signal to require consent.   
+* Restriction signals with respect to applicable legal basis for purposes for which a vendor has not declared flexibility may be ignored by vendors
```
alextcone, line 206: I feel like this bullet does not sufficiently achieve the aim of "avoidance of doubt". I under stand what you're getting at I think, but I think this bullet can either be struck or it should have a qualifier like "...who will determine whether or not processing is allowed based on x, y, z."

TCFv2/TCF-Implementation-Guidelines.md
```diff
@@ -251,7 +252,19 @@ You’ll usually want to go with a combination of server-side storage – for be
...
+
+    1. The vendor registered a purpose as legitimate interest (default legal basis), but also registered this purpose as flexible (i.e. also accepts consent as a legal basis). In this case a "require consent" restriction is needed to signal that the vendor may only apply the consent signal/can rely on it to process personal data.
```
Facens, line 262:
```suggestion
    1. The vendor registered a purpose as legitimate interest (default legal basis), but also registered this purpose as flexible (i.e. also accepts consent as a legal basis). In this case a "require consent" restriction is needed to signal that the vendor may only process using consent as legal basis.
```

TCFv2/TCF-Implementation-Guidelines.md
```diff
@@ -251,7 +252,19 @@ You’ll usually want to go with a combination of server-side storage – for be
...
+    2. The vendor registered a purpose as consent (default legal basis), but also registered this purpose as flexible (i.e. also accepts legitimate interest as a legal basis).  In this case a "require legitimate interest" restriction is needed to signal that the vendor may only apply the legitimate interest signal/can rely on it to process personal data.
```
Facens, line 263:
```suggestion
    2. The vendor registered a purpose as consent (default legal basis), but also registered this purpose as flexible (i.e. also accepts legitimate interest as a legal basis).  In this case a "require legitimate interest" restriction is needed to signal that the vendor may only process using legitimate interest as legal basis.
```

TCFv2/TCF-Implementation-Guidelines.md
```diff
@@ -251,7 +252,19 @@ You’ll usually want to go with a combination of server-side storage – for be
 ## 4. Withdrawal of consent and other non-TCF policy <a name="withdrawconsent"></a>
 Signals sent through the IAB Europe framework should only indicate what the user status is at the time of the signal creation and nothing else. While the CMP should also enable users to withdraw consent, the minimum requirement is to record the user's preference at the time the signal is created. Certain GDPR policy, such as portability and the right to be forgotten, is not covered in the IAB Europe TCF. CMPs and vendors should address other GDPR rights outside the TCF separately and on their own.
 
-## 5. CMP interface requirements<a name="cmpreq"></a>
+## 5. Encoding publisher restrictions <a name="pubrestrenc"></a>
```
alextcone, line 255: I am happy with this section. I think it does the job we're aiming to do in clarifying how a CMP should operate here.

TCFv2/TCF-Implementation-Guidelines.md
```diff
@@ -251,7 +252,19 @@ You’ll usually want to go with a combination of server-side storage – for be
...
+
+    1. The vendor registered a purpose as legitimate interest (default legal basis), but also registered this purpose as flexible (i.e. also accepts consent as a legal basis). In this case a "require consent" restriction is needed to signal that the vendor may only apply the consent signal/can rely on it to process personal data.
+    2. The vendor registered a purpose as consent (default legal basis), but also registered this purpose as flexible (i.e. also accepts legitimate interest as a legal basis).  In this case a "require legitimate interest" restriction is needed to signal that the vendor may only apply the legitimate interest signal/can rely on it to process personal data.
+    
+Note that the above does not preclude the use of efficient encoding/decoding schemes in certain scenarios, for example in cases where a publisher wants to restrict a purpose for all vendors to consent it might be more efficient to encode a small number of range restriction segments using a specific encoding scheme.
```
Facens, line 265:
```suggestion
Note that the above does not preclude the use of efficient encoding/decoding schemes in certain scenarios. For example, in cases where a publisher wants to restrict a purpose for all vendors to consent, it might be more efficient to encode a small number of range restriction segments using a specific encoding scheme.
```

alextcone (commented):
I only have one comment to be resolved. Otherwise I am happy with this and appreciate the work you've put into this.

Facens (changes requested):

TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
Facens: @alextcone not sure I follow, could you explain? The gist here is:
- If you declare a certain purpose under "purposes", thus with "consent" as your default legal basis, you can only fall back to legitimate interest for that purpose if you had declared it as flexible and if you're getting "2" for that purpose under restrictions {}
- If you declare a certain purpose under "legIntPurposes", thus with "legitimate interest" as your default legal basis, you can only fall back to consent for that purpose if you had declared it as flexible and if you're getting "1" for that purpose under restrictions {}
Maybe it's clearer if we put it this way? Or we could just add the above for EXTRA clarity.

alextcone: Anything that says a vendor can "ignore" something that isn't further qualified to say "but you can only ignore this part" makes me nervous about potential interpretations. This last bullet is the one I am concerned about:

> Restriction signals with respect to applicable legal basis for purposes for which a vendor has not declared flexibility may be ignored by vendors

Facens: I don't think that part is instrumental, it's more like a deduction, we could also remove it.
@asr-enid?

alextcone: I think if you're ok with removing it I am too. The explainers above make it clear in my mind. If we're going to keep it though we need to finish the thought.

asr-enid: I think @alextcone is only refering to 

> Restriction signals with respect to applicable legal basis for purposes for which a vendor has not declared flexibility may be ignored by vendors

We can strike that bullet no objection from my end

asr-enid: Ok with removing, I added this with a different topic in mind which is not relevant here. Will take care of it

asr-enid: See commit https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/pull/244/commits/d74f1fad0468c81c616bb06f40f6c310c8245b77

Facens:
@asr-enid Take a look at my comments too, all minor tweaks in English, my bad for not creating a review and filing them as single comments, thought I could convert them but can't find a way. Lmk!

TCFv2/TCF-Implementation-Guidelines.md
asr-enid: Would want to leave it like that, sentence one is already defining that the vendor was not disclosed, thus 2 picks of on that (In case...., If.... is double conditional here)

asr-enid:
I think we are good now, also applied the editorial comments (all but one) in https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/pull/244/commits/30f1686167747549e624380110ecc1d4412ca91f

Facens:
I think we're good! @alextcone FF to give it a final read and merge!

alextcone (approved):
Looks good. Thank you! Merging now.

PR #246, opened 2020-10-10
Fix incorrect statement about getTCData

bloodyowl:
This statement implies that getTCData callback is guaranteed to run
synchronously no matter the state of the __tcfapi function, which contradicts the following sentence:

> the only command required to be on the page in a stub before the rest of the commands are implemented

that we can read in the ping documentation.


alextcone:
Requesting reviewers. Just added @a2intl, but am also inviting @dmdabbs and @HeinzBaumann to the project

HeinzBaumann (approved):
This clarification is okay.

PR #251, opened 2020-11-11
Clarify macro usage

alextcone:
Based on discussion captured here: https://docs.google.com/document/d/1oL6qjoWSw69oOfQT_xdKh0HUNFfdE5piQ8nbgMMiQo8/edit#

### DIFF
--- a/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
+++ b/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
@@ -361,14 +361,19 @@ The following table sets expectations about which parties are expected to add an
    <td>DSP before returning creative URL to page</td>
   </tr>
   <tr>
-   <td>3rd party creative <strong>via publisher ad server</strong></td>
+   <td>3rd party creative URL <strong>via publisher ad server</strong></td>
    <td>Creative vendor at the time of tag export</td>
    <td>Publisher ad server before returning URL to the page</td>
   </tr>
   <tr>
-   <td>3rd party creative <strong>via header bidding</strong></td>
+   <td>3rd party creative URL <strong>via header bidding</strong></td>
    <td>Creative vendor at the time of tag export</td>
    <td>DSP before returning URL to the SSP</td>
+  </tr>
+  <tr>
+   <td>3rd party creative URL <strong>via mediation</strong></td>
+   <td>Creative vendor at the time of tag export</td>
+   <td>Mediating network before returning URL to the SSP</td>
   </tr>
   <tr>
    <td>Hosted creative URL <strong>via DSP</strong></td>
@@ -376,22 +381,22 @@ The following table sets expectations about which parties are expected to add an
    <td>DSP before returning creative URL to page</td>
   </tr>
   <tr>
-   <td>Hosted creative <strong>via publisher ad server</strong></td>
+   <td>Hosted creative URL <strong>via publisher ad server</strong></td>
    <td>Publisher ad server before returning URL to the page</td>
    <td>Publisher ad server before returning URL to the page</td>
   </tr>
   <tr>
-   <td>Hosted creative <strong>via header bidding</strong></td>
+   <td>Hosted creative URL <strong>via header bidding</strong></td>
    <td>DSP before returning URL to the SSP</td>
    <td>DSP before returning URL to the SSP</td>
   </tr>
   <tr>
-   <td>Cookie Sync Pixel <strong>via Creative</strong></td>
+   <td>Cookie Sync URL <strong>via creative</strong></td>
    <td>?</td>
    <td>?</td>
   </tr>
   <tr>
-   <td>Cookie Sync Pixel <strong>via Other Means</strong></td>
+   <td>Cookie Sync URL <strong>via other means</strong></td>
    <td>?</td>
    <td>?</td>
   </tr>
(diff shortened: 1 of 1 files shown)


TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
```diff
@@ -335,7 +297,153 @@ The service making the call must replace the macros with appropriate values desc
...
+  </tr>
+  <tr>
+   <td>Pixels added to creatives</td>
+   <td>Pixel vendor</td>
+   <td>Vendor calling the initial creative URL?</td>
+  </tr>
+  <tr>
+   <td>URL pixels placed on digital properties</td>
+   <td>Pixel vendor</td>
+   <td>INSERT</td>
```
williamleeMM, line 438: I think this varies based on how a digital property is set up, so I think here it would be dependent on the digital property to themselves expand the macros provided to them by vendors.

TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
```diff
@@ -335,7 +297,153 @@ The service making the call must replace the macros with appropriate values desc
...
+   <td>DSP before returning URL to the SSP</td>
+   <td>DSP before returning URL to the SSP</td>
+  </tr>
+  <tr>
+   <td>Cookie Sync URL <strong>via creative</strong></td>
+   <td>INSERT</td>
+   <td>INSERT</td>
+  </tr>
+  <tr>
+   <td>Cookie Sync URL <strong>via other means</strong></td>
```
williamleeMM, line 426: There’s likely a need for multiple macros (provided and expanded by multiple parties) here. For example:

- where company_1 (vendor id 123) and company_2 (vendor id has 987) cookie sync; and
- company_1 is initiating the cookie sync with company_2

company_2 may provide company_1 with tag containing a url to be placed on one of their partner’s digital properties to initiate that cookie sync - e.g.:

`<script type="text/javascript" src="//pixel.company_2.com/sync/js?gdpr=[GDPR]&has gdpr_consent=[GDPR_CONSENT_987]&return_url=[RETURN_URL]></script>`

The GDPR and GDPR_CONSENT_987 macros would be expanded by the digital property the tag was placed on by company_1. The RETURN_URL macro in this case should be expanded by company_1, with the url where they would like for company_2 to direct any cookie syncs to and appropriate macros. For example, it could be expanded with “https://sync.company_1.com/company_2?user_id=[USER_ID]&gdpr=[GDPR]&gdpr_consent=[GDPR_CONSENT_123]” - where USER_ID, GDPR, and GDPR_CONSENT_123 are macros to be expanded by company_2.

williamleeMM:
Thanks for putting this together @alextcone!

I think it would make sense to clarify that where vendors provide tags (of any kind - img, js, or otherwise) with appropriate macros to their digital property partners to call from their pages, it is the digital property’s responsibility to expand those macros where they deem it to be appropriate.

For example a vendor could provide a digital property partner with either of the following tags:

``

`<script language='JavaScript1.1' async src='//pixel.vendorcom/event/js?GDPR=${GDPR}&GDPR_CONSENT=${GDPR_CONSENT_79}'></script>`

Although the latter tag is a script tag, as appropriate macros are provided, the digital property should populate the macros.

TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
```diff
@@ -335,7 +297,153 @@ The service making the call must replace the macros with appropriate values desc
...
+    </tr>
+    <tr>
+      <td><code>gdpr_pd</code></td>
+      <td><code>${GDPR_PD}</code></td>
+      <td><code>&gdpr_pd=${GDPR_PD}</code></td>
+    </tr>
+  </tbody>
+</table>
+
+The vendor service making the call to a URL with the macro must replace the macros with appropriate values described in the table below. For macro `${GDPR_CONSENT_XXXXX}`, the vendor service making the call to the URL must also check that the macro contains a valid Vendor ID and that the Vendor ID provided has a legal basis for processing personal data _before_ replacing the macro and calling the URL.
```
asom213, line 352: Hi @alextcone ,

If the callee provided URL didn’t include ${GDPR_CONSENT_XXXXX} macros, should we consider the URL an out-of-TCF use case? If so, should the caller stops calling the URL?

TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
```diff
@@ -335,7 +297,153 @@ The service making the call must replace the macros with appropriate values desc
   </tbody>
 </table>
 
-**Note:** other personal data, like IP addresses or callee cookies, may be passed as part of the request, and the `gdpr` and `gdpr_consent_xxxxx` is used by the callee to determine whether an identifier cookie or other personal data can be set and/or used.
+**Note:** other personal data, like IP addresses or callee cookies, may be passed as part of the request, and the `${GDPR}` and `$GDPR_CONSENT_XXXXX` is used by the callee to determine whether an identifier cookie or other personal data can be set and/or used.
```
simontrasler, line 300: Is this a typo and should read `${GDPR_CONSENT_XXXX}`?

paulandrewfarrow (commented):
I feel like the whole section starts from the point of view of macros, when really it should start with the params. The macros are a means to provide the required parameters, but it reads a little like it's the other way around. 

I personally think it would be easier to read if it followed this structure:

1. Vendors utilising URL-based ad requests need to accept gdpr parameters via the URL query string.
2. These params are as follows...
3. Vendors should modify their URL-based ad requests to contain these parameters as well as macro placeholders for parameters.
4. Those macros are as follows...
5. Here is an example of a modified URL ad request:https://guides.github.com/features/mastering-markdown/
6. Supply-chain members with access to the gdpr parameters need to replace these macros before making the URL-based ad request. Examples of members could be:

- Website owners
- Downstream Vendors

7. Website owners expand the macros directly by reading the CMP API on the page
8. Downstream Vendors (SSPs/ad servers/DSPs/etc.) need to expand these macros when receiving them in their own requests, and then returning responses/forwarding requests with gdpr parameters filled out. Use cases might include:

- Delivery of Creative containing pixel (explain use case)
- ...

TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
lon-pilot-mpf: @asom213 I think that's something you need to discuss with your legal team, but the policies do state that (Chapter 3 16.19): "A Vendor must not transmit a user’s personal data to an entity outside of the Framework unless it has a justified basis for relying on that entity’s having a Legal Basis for processing the personal data in question."

PR #254, opened 2020-11-17
Implement Accept All/Reject All buttons in reference implementation

antoine-g:
Based on the guidance provided by the ICO (UK) at
https://ico.org.uk/for-organisations/guide-to-pecr/guidance-on-the-use-of-cookies-and-similar-technologies/how-do-we-comply-with-the-cookie-rules/#comply12

> Depending on the circumstances, particularly the design of your
consent mechanism and the wording you use in the information you
provide, it is also likely that predetermining non-essential cookies
could be considered as ‘nudge behaviour’ – ie, you are influencing the
user to take a particular course of action. 

> A consent mechanism that emphasises ‘agree’ or ‘allow’ over ‘reject’
or ‘block’ represents a non-compliant approach, as the online service is
influencing users towards the ‘accept’ option.

> A consent mechanism that doesn’t allow a user to make a choice would
also be non-compliant, even where the controls are located in a ‘more
information’ section.

Also based on the guidance provided by the CNIL (FR) with an excerpt
translated below. See
https://www.cnil.fr/sites/default/files/atoms/files/recommandation-cookies-et-autres-traceurs.pdf
for the full document.

> 30. The data controller must offer users both the possibility of
accepting and refusing read and/or write operations with the same degree
of simplicity.

> 31.  The Commission therefore strongly recommends that the mechanism
for expressing a refusal to consent to read and/or write operations
should be accessible on the same screen and with the same degree of ease
as the mechanism for expressing consent. Indeed, it believes that
consent collection interfaces that require a single click to consent to
tracing while several actions are required to "set up" a refusal to
consent present, in most cases, the risk of biasing the user's choice,
who wishes to be able to view the site or use the application quickly.

> For example, at the first level of information, users may have a
choice between two buttons presented at the same level and in the same
format, with "accept all" and "refuse all", "allow" and "prohibit", or
"consent" and "do not consent", or other equivalent and sufficiently
clear wording, respectively.  The Commission considers that this
modality constitutes a simple and clear means of enabling the user to
express his refusal as easily as his consent.

> Figure 4 - The user can choose between an "accept all" and a "refuse
all" button presented at the same level and in the same format.

Screenshot of the new reference implementation with the changes made as part of this pull request:

Screenshot of modified reference implementation

### DIFF
--- a/reference/src/components/popup/intro/intro.jsx
+++ b/reference/src/components/popup/intro/intro.jsx
@@ -21,6 +21,7 @@ export default class Intro extends Component {
 
 		const {
 			onAcceptAll,
+			onRejectAll,
 			onShowPurposes,
 			onClose
 		} = props;
@@ -46,10 +47,16 @@ export default class Intro extends Component {
 						<LocalLabel localizeKey='showPurposes'>Manage your choices</LocalLabel>
 					</Button>
 					<Button
-						class={style.acceptAll}
+						class={style.choiceButton}
 						onClick={onAcceptAll}
 					>
-						<LocalLabel localizeKey='acceptAll'>Got it, thanks!</LocalLabel>
+						<LocalLabel localizeKey='acceptAll'>Accept All</LocalLabel>
+					</Button>
+					<Button
+						class={style.choiceButton}
+						onClick={onRejectAll}
+					>
+						<LocalLabel localizeKey='RejectAll'>Reject All</LocalLabel>
 					</Button>
 				</div>
 			</div>
--- a/reference/src/components/popup/intro/intro.less
+++ b/reference/src/components/popup/intro/intro.less
@@ -15,17 +15,22 @@ div.intro {
 		margin: 4% 0;
 		overflow: auto;
 	}
+
 	.options {
 		display: flex;
 		flex-shrink: 0;
 		width: 100%;
 		padding-bottom: 10px;
+
+		button {
+			margin-right: 10px;
+		}
+		div:first-child{
+			margin-left: 0;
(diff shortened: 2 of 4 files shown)


MathRobin:
Please, if this PR is good, merge it. I ask it to you as the developer of browser extension Never-Consent.

Just a quick reminder, next April 1st 2021 (in three days so), french CNIL consider that 

Sources:
- https://www.usine-digitale.fr/editorial/consentement-banniere-finalites-ce-qui-va-changer-au-1er-avril-en-matiere-de-cookies-publicitaires.N1076564
- https://www.sudouest.fr/france/internet-fin-du-sursis-pour-les-cookies-la-publicite-sur-le-web-en-pleine-reconversion-1833771.php

> 31. The data controller must offer users both the possibility of
>    accepting and refusing read and/or write operations with the same degree
>    of simplicity.

So all your users will be eligible to legal proceedings. Have fun^^

antoine-g:
It seems that the reference UI implementation has been removed from the repository and left to be implemented by the CMP providers, shifting this way the liability to be compliant with the regulations.

Today, the French regulator has fined both Google and Facebook for the absence of top level rejection option as part of the implementation of their consent screens. See https://www.cnil.fr/en/cookies-cnil-fines-google-total-150-million-euros-and-facebook-60-million-euros-non-compliance.

A comment, update or status post to this pull request from the maintainers would have been appreciated given that it has been opened more than a year ago.

MathRobin:
You hope that people will be honest? They work for marketing. Off course they will not be trustful.

PR #261, opened 2021-02-05
clearly handle no cookie use

alextcone:
Currently, IAB Europe, the managing organization (MO) for TCF is using the value -100 as a proxy to indicate a GVL vendor does not use cookies of any kind (including session cookies). This pull request is meant to add a clear indication to CMPs and subsequently to users that a vendor does not use cookies of any kind.

### DIFF
--- a/TCFv2/IAB Tech Lab - Device storage duration and access disclosure.md
+++ b/TCFv2/IAB Tech Lab - Device storage duration and access disclosure.md
@@ -67,7 +67,7 @@ The number of seconds representing the longest potential duration for cookie sto
   <tr>
    <td><code>cookieMaxAgeSeconds</code>
    </td>
-   <td><strong>optional</strong>
+   <td><strong>Required if useCookies is set to True, else optional</strong>
    </td>
    <td>integer
    </td>
--- a/TCFv2/IAB Tech Lab - Device storage duration and access disclosure.md
+++ b/TCFv2/IAB Tech Lab - Device storage duration and access disclosure.md
@@ -16,6 +16,37 @@ In its Planet49 judgment (available [here](http://curia.europa.eu/juris/document
 
 ## Global Vendor List Fields<sup id="a2">[2](#f2)</sup>
 
+### <code>usesCookies</code>
+
+This true or false field indicates whether the vendor uses cookie storage (session or otherwise).
+
+<table>
+  <tr>
+   <td>Field
+   </td>
+   <td>Scope
+   </td>
+   <td>Type
+   </td>
+   <td>Default
+   </td>
+   <td>Description
+   </td>
+  </tr>
+  <tr>
+   <td><code>usesCookies</code>
+   </td>
+   <td><strong>required</strong>
+   </td>
+   <td>Boolean
+   </td>
+   <td>-
+   </td>
+   <td>Indicates whether the vendor uses cookie storage (session or otherwise). True indicates cookie storage is used. False cookie storage is not used. 
+   </td>
+  </tr>
+</table>
+
 ### <code>cookieMaxAgeSeconds</code>
 
 The number of seconds representing the longest potential duration for cookie storage on a device. If a Vendor uses multiple cookies with differing durations, <code>cookieMaxAgeSeconds</code> represents the cookie with the longest duration. Note: cookies are the only method of storage or device access that permit a predictable duration to be set. 
@@ -36,7 +67,7 @@ The number of seconds representing the longest potential duration for cookie sto
   <tr>
    <td><code>cookieMaxAgeSeconds</code>
    </td>
-   <td><strong>required</strong>
+   <td><strong>optional</strong>
    </td>
    <td>integer
    </td>
@@ -73,7 +104,7 @@ This true or false field indicates whether the vendor uses other, non-cookie met
    </td>
    <td>-
    </td>
-   <td>Indicates the vendor’s use of non-cookie storage and access to information already stored on a user’s device. True indicates non-cookie access is used. False indicates only cookie storage and access are used. 
+   <td>Indicates the vendor’s use of non-cookie storage and access to information already stored on a user’s device. True indicates non-cookie access is used. False indicates non-cookie storage and access to information already stored on a user's device <em>is not</em> used. 
    </td>
   </tr>
 </table>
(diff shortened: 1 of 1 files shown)


janwinkler (approved):

HeinzBaumann:
I think this looks good. I do have a question though. I see that cookieMaxAgeSeconds is not optional. Is this now optional in all cases or does it need to be set when usesCookies is set to true?  Do we need to clarify that in the documentation?

HeinzBaumann:
Suggestion was made by Anne Goodman:
Anne Goodman
3:12 AM (11 hours ago)
to alex.cone@iabtechlab.com, me

Hi there,

In preparation for the call covering off the point that Heinz you raise on GitHub ref pt 5 below (see highlighted)

Heinz comment
I think this looks good. I do have a question though. I see that cookieMaxAgeSeconds is not optional. Is this now optional in all cases or does it need to be set when usesCookies is set to true? Do we need to clarify that in the documentation?

The GVL currently asks registering vendors to indicate 

Not applicable - no cookies used

The completion of this field would drive the outcome usesCookies set false and at the same time field cookieMaxAgeSeconds* is no longer required. If they did not complete the field usesCookies then this field would be set to true and the cookieMaxAgeSeconds* is required.

Therefore the cookieMaxAgeSeconds* is not required if usesCookies set false

Should the documentation that currently reads "required" and "optional" actually read "required" and "not required" (therefore making the documentation clearer reflecting on Heinz point)?

<td><strong>required</strong> 
<td><strong>optional</strong>

Anne Goodman @aol.com 
07802 784418

TCFv2/IAB Tech Lab - Device storage duration and access disclosure.md
```diff
@@ -36,7 +67,7 @@ The number of seconds representing the longest potential duration for cookie sto
   <tr>
    <td><code>cookieMaxAgeSeconds</code>
    </td>
-   <td><strong>required</strong>
+   <td><strong>optional</strong>
```
lon-pilot-mpf, line 70: Given Anne's comment, perhaps "Required if useCookies is set to True, else optional" is better here?

tla-sirdata:
> A negative number or a 0 indicate session storage similar to the Set-Cookie spec

Should we update that line since a negative number is no longer possible ?

TCFv2/IAB Tech Lab - Device storage duration and access disclosure.md
HeinzBaumann: Agreed. @alextcone can you do that change in the doc, please?

HeinzBaumann (changes requested):

TCFv2/IAB Tech Lab - Device storage duration and access disclosure.md
BorisSambil: I think it’s definitely good to have that extra attribute usesCookies true|false and only integers for cookieMaxAgeSeconds. But that raises following question for me: The MO officially announced that cookieMaxAgeSeconds will be null in the gvl for the case when the vendor checked „I don’t use cookies“ (and postponed the gvl change until March 11th) - If we go with this PR and make the cookieMaxAgeSeconds optional - I dont see that the null value makes sense. (And don’t think the gvl change should happen if we shortterm apply this pr)

alextcone: @HeinzBaumann and @lon-pilot-mpf, resolving now.

HeinzBaumann (approved):

HeinzBaumann (approved):

lon-pilot-mpf (approved):

dmdabbs (approved):

alextcone:
> > A negative number or a 0 indicate session storage similar to the Set-Cookie spec
> 
> Should we update that line since a negative number is no longer possible ?
> 
> `maxAgeSeconds` should be also clarified in the disclosure object for session cookies.

@tla-sirdata - done

lon-pilot-mpf:
Hey @alextcone could you please merge master into this branch? It is a bit confusing as certain features of the master branch (e.g `cookieRefresh`) are missing.

TCFv2/IAB Tech Lab - Device storage duration and access disclosure.md
```diff
@@ -36,13 +67,13 @@ The number of seconds representing the longest potential duration for cookie sto
   <tr>
    <td><code>cookieMaxAgeSeconds</code>
    </td>
-   <td><strong>required</strong>
+   <td><strong>required if useCookies is set to true, else optional</strong>
```
Facens, line 70:
```suggestion
   <td><strong>required if usesCookies is set to true, else optional</strong>
```

Facens (changes requested):
IAB Europe is now using 'cookieMaxAgeSeconds: null' as an indication that cookies are not in use. Could you please clarify why we feel the need to add the usesCookies field? I'm not against in principle, but it would be a proxy of 'cookieMaxAgeSeconds: null' and it has the drawback of increasing the size of the GVL.

alextcone:
> IAB Europe is now using 'cookieMaxAgeSeconds: null' as an indicator that cookies are not used. Could you please clarify why we feel the need to add the usesCookies field? I'm not against it in principle, but it would be a proxy of 'cookieMaxAgeSeconds: null' and it has the drawback of increasing the size of the GVL.

IAB Europe is now aligned with the change that makes this a clear, new field instead of using null or a negative number on a different field to mean the same thing. From a note this morning Anne Goodman at IAB Europe sent:

> From: Anne Goodman
> Reply-To: Anne Goodman 
> Date: Tuesday, March 9, 2021 at 1:54 AM
> To: Alex Cone
> Subject: Re: GVL field CookiesMaxAge - FINAL
> ...
> 
> TO BE CLEAR - I will update the GVL JSON so that it reflects the commit on usesCookies and will be populated with the additional field in the GVL based on vendors checking the "Not applicable - no cookies used" (true) or not (false) in their registration. When this field is set to true, the cookieMaxAgeSeconds field be NULL instead of being set to -100 by default.
> 
> The above has been actioned and should be live on 11th Thursday. 
> 
> ...
> 
> Thanks
> Anne

PR #273, opened 2021-04-12
initial special purposes only vendors commit

alextcone:
Handling transparency for vendors who are only signed up for Special Purposes

### DIFF
--- a/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
+++ b/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
@@ -65,6 +65,7 @@
 
 | Date | Version | Comments |
 | :-- | :-- | :-- |
+| April 2021 | 2.0 | Special Purpose only vendors transparency |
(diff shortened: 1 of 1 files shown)


anderagakura:
@alextcone After have discussed internally below is the description we would like to push : 

`Vendors that register for Special Purposes only, and no other Purposes, and have been displayed by the CMP, then the value must be set to <code>1</code>. 
Note: Special Purposes are displayed for transparency only and do not enable user choice. `

a2intl:
The "then the value must be set" is incorrect grammar, it should be "should have their value set".

anderagakura:
> The "then the value must be set" is incorrect grammar, it should be "should have their value set".

@a2intl We use "must" because we want to indicate this case is **required** for the vendor’s bit in order to display it for transparency.

alextcone:
Fixed grammar (for @a2intl) while keeping "must" per @anderagakura's last comment. Ready to merge.

PR #274, opened 2021-04-14

dmdabbs:
Opening a PR in support of this TCFlib PR: https://github.com/InteractiveAdvertisingBureau/iabtcf-es/pull/262.

### DIFF
--- a/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
+++ b/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
@@ -1134,7 +1134,7 @@ CLcVDxRMWfGmWAVAHCENAXCkAKDAADnAABRgA5mdfCKZuYJez-NQm0TBMYA4oCAAGQYIAAAAAAEAIAEg
       <td>
         The last ID of the inclusive contiguous ascending-order series of
         Vendor IDs started with StartOrOnlyVendorId but only if that series
-        has a cardinality greater than 1, otherwise this field is omitted.
+        has a cardinality greater than 1, otherwise this field is omitted. Note that _contiguous_ above permits encoding ranges that cover gaps in the GVL registrations unlike other range encodings in this specification. For example, the publisher asks the CMP to signal a Publisher Restriction on Purpose 2 on Vendors [12, 15, 18, 23]. In the current GVL, Vendors 13-14, 16-17 and 19-22 are not registered. It is permitted to encode a single range sub-entry spanning 12-23 versus a separete entry for each restricted Vendor. 
       </td>
       <td>
         The end of the series of Vendor IDs – this is omitted if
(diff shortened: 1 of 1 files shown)


tla-sirdata (approved):
Looks good to me. Thanks David!

a2intl (approved):

TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
```diff
@@ -1134,7 +1134,7 @@ CLcVDxRMWfGmWAVAHCENAXCkAKDAADnAABRgA5mdfCKZuYJez-NQm0TBMYA4oCAAGQYIAAAAAAEAIAEg
       <td>
         The last ID of the inclusive contiguous ascending-order series of
         Vendor IDs started with StartOrOnlyVendorId but only if that series
-        has a cardinality greater than 1, otherwise this field is omitted.
+        has a cardinality greater than 1, otherwise this field is omitted. Note that _contiguous_ above permits encoding ranges that cover gaps in the GVL registrations unlike other range encodings in this specification. For example, the publisher asks the CMP to signal a Publisher Restriction on Purpose 2 on Vendors [12, 15, 18, 23]. In the current GVL, Vendors 13-14, 16-17 and 19-22 are not registered. It is permitted to encode a single range sub-entry spanning 12-23 versus a separate entry for each restricted Vendor.
```
a2intl, line 1137: nit: I think we can leave out many of these words, like: `The publisher asks the CMP to signal a Publisher Restriction on Purpose 2` because we're already in that context. Perhaps just `For example, to encode a restriction for vendors [12, 15, 18, 23] if vendor IDs 13-14, 16-17, and 19-22 are not in the GVL, a range entry of 12-23 is permitted.`.

a2intl: Another nit: we may want to be explicit about deleted vendorId handling / allowability.

a2intl: I would also not use the word "registrations" (as it's not 100% clear what that's referring to), and speak only about vendor IDs that are being encoded or are in the GVL.

dmdabbs: Great feedback, @a2intl  I've amended to use your suggestion.

dmdabbs: >Another nit: we may want to be explicit about deleted vendorId handling / allowability.
Does the PR look at deleteDate? I'm not as familiar how the lib data model handles GVL vendors with non-empty deletedDates. If it does not include them in the data presented to the CMP, then they are the same as where no Vendir is registered for that ID. 

The Framework Signals discussion briefly touched on deleted use case and that it would also be innocuous to include those 'gaps. '

dmdabbs: How about this to cover bases on 'deleted' vendors:

Note that _contiguous_ above permits encoding ranges that include deleted Vendors or cover gaps in GVL vendor IDs unlike other range encodings in this specification. For example, to encode a restriction for vendors [12, 15, 18, 24] if vendor IDs 13-14, 16-17, and 19-22 are not in the GVL and 23 has a non-empty _deletedDate_, a range entry of 12-24 is permitted.

a2intl (approved):

PR #300, opened 2021-08-28
Adding optional use of Disclosed Vendors / Bridging for potential signaling re-use

asr-enid:
PR w.r.t. the the changes discussed on global privacy wg call 8/26/21

- Adding optional use of _DisclosedVendores_ while storing service level TC Strings by CMPs with reference to reference implementation support, thus giving it a scope beyond OOB outside of signalling
- No changes in signaling or vendor processing (optional "internal" use for CMPs)
- Bridging for potential future re-use in signaling using this or an adapted segment


TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
```diff
@@ -1202,41 +1203,42 @@ COvFyGBOvFyGBAbAAAENAPCAAOAAAAAAAAAAAEEUACCKAAA
...
+The _**DisclosedVendors**_ is a TC String segment that signals which vendors have been disclosed to a given user by a CMP. This segment is required when saving a global-context TC String.  When a CMP updates a globally-scoped TC String, the CMP <span style="text-decoration:underline;">MUST</span> retain the existing values and only add new disclosed Vendor IDs that had not been added by other CMPs in prior interactions with this user. 
 
+The _**DisclosedVendors**_ segment may also be used by a CMP while [storing](#how-should-a-transparency--consent-string-be-stored) service-specific TC Strings to retain the vendors that have been disclosed to a user by a given CMP on a given site/app or group of sites/apps. For the avoidance of doubt: The use of the _**DisclosedVendors**_ segment is only permissible in relation to service-specific TC Strings with respect to storing it as additional segment. CMPs <span style="text-decoration:underline;">MUST</span> ensure that the segment is not used when signaling to vendors. The official tool suite supports this through [encoding options](#https://github.com/InteractiveAdvertisingBureau/iabtcf-es/tree/master/modules/core#encoding-options).
```
a2intl, line 1241: I'd replace "signaling to vendors" with the more technically-specific phrasing used elsewhere: "When a TC String is requested from (or returned by) the CMP API it shall..." As this is a technical spec, we can just say what the requirements are, once and briefly, rather than using "For the avoidance of doubt" and repetition for emphasis. Perhaps replace the first two sentences with:

> The _**DisclosedVendors**_ segment may also be used by a CMP while [storing](#how-should-a-transparency--consent-string-be-stored) service-specific TC Strings to record which vendors have been disclosed to the user, but must not be included in the TC String when returned by the CMP API.

It also seems pecuiliar that the spec would "permit storage" of any of the segments in service-specific storage. Since the service-specific storage is neither specified nor controlled, and completely up to the CMP, they can put anything they want in the storage, in whatever format they choose, as long as they return compliant strings via the API. Perhaps we should avoid making any requirements/permittances as to service-specific storage whatsoever, since it should not (and can not) matter.

TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
```diff
@@ -184,7 +185,7 @@ A TC String that contains positive consent signals must not be created before cl
...
-*   **Service-specific** - A  TC String in this context is only used by the site(s) or app(s) on which it is running. One is created for every user on a given site/app or group of sites/apps. They may contain [Publisher restrictions](#what-are-publisher-restrictions), a _**[Publisher TC](#publisher-purposes-transparency-and-consent)**_ segment and an _**[AllowedVendors](#allowed-vendors-oob)**_ segment.
+*   **Service-specific** - A  TC String in this context is only used by the site(s) or app(s) on which it is running. One is created for every user on a given site/app or group of sites/apps. They may contain [Publisher restrictions](#what-are-publisher-restrictions), a _**[Publisher TC](#publisher-purposes-transparency-and-consent)**_ segment and an _**[AllowedVendors](#allowed-vendors-oob)**_ segment when signaling to vendors. The use of the _**[DisclosedVendors](#disclosed-vendors)**_ segment is restricted to storing it as additional metadata with the service-specific TC String.
```
a2intl, line 188: again, I think we should replace the phrase "signaling to vendors" with "returned by the CMP API", and perhaps avoid any mention of what a CMP can or cannot put into service-specific storage.

TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
asr-enid: The suggestion works for me - thanks!

I used signaling given its also used throughout the document aside from returned by the CMP API. No need to use" permit storage" here language wise.

TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
asr-enid: Happy to replace "_when signaling to vendors_" with "_returned by the CMP API_" and remove the last sentence.

alextcone (approved):
Thank you @asr-enid. I think this looks good. Agree with @a2intl's suggestions which it appears you've incorporated.

asr-enid:
@alextcone suggested changes have been added to the PR just now

a2intl (approved):

alextcone:
@anderagakura - I am merging this much smaller change in ahead of the Global Scope/OOB deprecation PR you have open at the moment (#301). I recommend pulling these #300 changes into #301. There are likely to be some conflicts as you've likely pulled out DisclosedVendors in #301 altogether while the approved #300 keeps DisclosedVendors for storage purposes only.

PR #301, opened 2021-09-02
TCF update: deprecate global scope, OOB

anderagakura:
Deprecation of Global Scope, OOB and euconsent-v2 cookie associated with the consensu.org domain

PR #303, opened 2021-09-09
Deprecate global scope, OOB

anderagakura:


TCFv2/TCF-Implementation-Guidelines.md
```diff
@@ -21,10 +21,9 @@ Policy FAQ, webinars, and other resources are available at
 &nbsp;&nbsp;&nbsp;&nbsp;**[Is v2 backwards compatible?](#compatibility)**<br>
 &nbsp;&nbsp;&nbsp;&nbsp;**[How do I evaluate the details provided in the TC String?](#evaluatetcstring)**<br>
 &nbsp;&nbsp;&nbsp;&nbsp;**[How should I handle multiple signals with different information?](#mergesignals)**<br>
-&nbsp;&nbsp;&nbsp;&nbsp;**[What is OOB?](#oob)**<br>
 ### [Publisher guidelines](#pub)
 &nbsp;&nbsp;&nbsp;&nbsp;**[What is a Consent Management Platform (CMP) and why do I, as a Publisher, need one?](#whatiscmp)**<br>
-&nbsp;&nbsp;&nbsp;&nbsp;**[What publisher controls are available? What happened to Pubvendors?](#pubvendor)**<br>
+&nbsp;&nbsp;&nbsp;&nbsp;**[What publisher controls are available? What happened to Pubvendors?](#pubvendors)**<br>
```
alextcone, line 26: If we're keeping "What happened to Pubvendors?" might we also consider adding something like "What happening to Global Scope and Out of Band (OOB)?"

TCFv2/TCF-Implementation-Guidelines.md
```diff
@@ -130,14 +125,11 @@ Under the GDPR, a legal basis is required for processing a user’s data. While
...
 
 ## How do I evaluate the details provided in the TC String?<a name="evaluatetcstring"></a>
-The TC String includes four (4) segments of information: the core string, publisher restrictions, publisher-approved vendors, and out-of-band (OOB) signaling. The technical specs describing the TC String provide details on specific information provided in each segment. These details may change from the start of the transaction to the end of the transaction. Vendors must evaluate the four segments of a string as it relates to a given transaction, determine the intent of the information provided, and proceed accordingly.
+The TC String returned by the CMP API can include (2) segments of information : the core string and the publisher TC segment." and removing altogether this part : "Vendors must evaluate the two segments of a string as it relates to a given transaction, determine the intent of the information provided, and proceed accordingly.
```
alextcone, line 128: It's possible people may be confused by the fact that the CMP API docs mention 3 segments (with the caveat that 1 of the segments is for storage purposes only). Am I over thinking this @asr-enid?

TCFv2/TCF-Implementation-Guidelines.md
AnneGoodman-IABEurope: Agree and propose the following text:
What happened to Global scope and Out-of-Band ?

The TCF Policy previously allowed legal bases in the Framework to be established with global scope, which means a legal basis is not only applicable on the service or group of services (service-specific and group-specific scopes) where it is obtained and managed, but all services implementing global scope preferences. In the context of global scope, TC Strings were stored in a 3rd party cookie associated with the “consensu.org” domain that enabled Consent Management Providers (CMPs) to read the “TC strings from their subdomains [cmp-name].mgr.consensu.org across different services. Deprecation of global-scope support was announced on June 22nd 2021. Alongside the deprecation of global scope, support for Out-of-Band (OOB) - which refers to instances where a legal basis has not been established using the TCF in a global-scope context - was also deprecated. Since Sept 1st 2021, TC strings established with global-scope are considered invalid."

alextcone:
Close in favor of #305 which combines latest #301, #303, and #304

PR #306, opened 2021-11-30
Reduce timestamp granularity

anderagakura:
While we keep using years, months and days, this update clears hours, minutes and seconds to 0.

### DIFF
--- a/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
+++ b/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
@@ -26,6 +26,7 @@
    + [How does a URL-based service process the TC string when it can't execute JavaScript?](#how-does-a-url-based-service-process-the-tc-string-when-it-cant-execute-javascript)
      - [Full TC String passing](#full-tc-string-passing)
    + [What if consent is governed differently in a country?](#what-if-consent-is-governed-differently-in-a-country)
+   + [What happened to Created and LastUpdated?](#what-happened-to-created-and-lastupdated)
  * [Creating a TC String](#creating-a-tc-string)
    + [How should a Transparency & Consent String be stored?](#how-should-a-transparency--consent-string-be-stored)
    + [What are the Purposes and Features being supported?](#what-are-the-purposes-and-features-being-supported)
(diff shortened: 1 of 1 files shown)


HeinzBaumann (approved):

TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
```diff
@@ -398,20 +404,19 @@ CLcVDxRMWfGmWAVAHCENAXCkAKDAADnAABRgA5mdfCKZuYJez-NQm0TBMYA4oCAAGQYIAAAAAAEAIAEg
...
       <td>36 bits</td>
       <td>
-        Epoch deciseconds when this TC String was first created (should not
-        be changed unless a new TCString is created from scratch)
+        Epoch time format when TC String was last updated (Must be updated any time a value is changed)
       </td>
       <td rowspan="2">
-        A decisecond is 1/10th of a second. To create a decisecond timestamp
-        in JavaScript: <code>Math.round((new Date()).getTime()/100)</code>
+       To create a timestamp in JavaScript: <code>Math.round((new Date(new Date().getFullYear(), new Date().getMonth(), new Date().getDate(), 0, 0, 0)).getTime()/100)</code>.
```
HeinzBaumann, line 410: I think I found a little issue with the timestamp.  Because we write out the timestamp in milliseconds since JAN 01 1970. (UTC) we will have the hour recorded in the timestamp even if we set the date to 0 hours 0 minutes 0 sec. The milliseconds we use is UTC. Each time zone will have a different millisecond value for each local 0h, 0m. 0s. Perhaps we need to make it explicit to use UTC time to set the timestamp. That way the millisecond value will be consistent across the time zones.

dmdabbs: Specifying uniform alignment to UTC is good @HeinzBaumann. How is this for the illustrative ECMAScript code:
```
var tmpDate = new Date();
var tcStampValue = Math.round((new Date(tmpDate.getUTCFullYear(), tmpDate.getUTCMonth(), tmpDate.getUTCDate(), 0, 0, 0)).getTime()/100);
```

dmdabbs: @anderagakura you'll need to amend the PR to also specify that the encoded datestamp value must be based on UTC time.

anderagakura: The change are applied [here](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/pull/306/commits/438ef34f336a6ccb3ecc61890f4c101811bf70b8)

HeinzBaumann (approved):

dmdabbs:
FWIW, the [community tclib DateEncoder](https://github.com/InteractiveAdvertisingBureau/iabtcf-es/blob/master/modules/core/src/encoder/field/DateEncoder.ts) uses `Date.getTime()` which expresses values in UTC, according to [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime)

```
getTime() always uses UTC for time representation. 
For example, a client browser in one timezone, getTime() will be the same as a client browser in any other timezone.
```

So I don't think users of the community tclib will need to make any special date-related code changes. We can change the lib to a) set `Created` to the `LastUpdated` value when reading existing strings and b) the converse when writing. Will look into it after holidays. 

cc: @shortaflip

dmdabbs (approved):

alextcone (approved):

PR #308, opened 2022-01-10
Device storage update

anderagakura:


dmdabbs:
Looks good to me, @anderagakura. 

Vendors may have integrations where a publisher or advertiser client grants access to implement code (and access storage) in a subdomain of the publisher or advertiser, such as TCF used to delegate to authorised CMPs for consensu.org. What is the guidance for disclosing these situations?  Surely the Vendor cannot publish a `Domains` object with '*' as the domain name.   Is this a case where it is incumbent on the Publisher or Advertiser (who own the domain) to ensure that they are a) disclosed in the CMP and b) this storage is disclosed with with proper Purpose mappings? The disclosure will not be associated with the Vendor partner, however.

tla-sirdata:
Looks good to me as well. Thanks

PR #312, opened 2022-02-01
Additional Vendor Information List

anderagakura:
This describes the B2B registration

### DIFF
--- a/TCFv2/Additional Vendor Information List Specification.md
+++ b/TCFv2/Additional Vendor Information List Specification.md
@@ -6,7 +6,7 @@ The Additional Vendor Information List is a technical document that can be downl
 
  * Full Legal Entity Address
  * B2B Contact details
- * Territorial scope (EU/EEA/UK countries where the vendor operates)
+ * Territorial scope (EU/EEA/EFTA/UK countries where the vendor operates)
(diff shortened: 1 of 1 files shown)


tla-sirdata:
Looks good to me. Thanks

janwinkler:
see my comments in google docs:
- CH is not in the EEA
- address should be formatted
- allow "EU" to use all countries instead of naming them all

anderagakura:
@janwinkler It's an EFTA country that has transposed the GDPR in national law and publishers there use the TCF. CH has not transposed the ePrivacy directive which is why the purposeOneTreatment flag is still relevant. The aim of this [push](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/pull/312/commits/64255a87f61b6e61686080072b8f304fdb73ef13) is to add EFTA territorial scope within the doc.

> CH is not in the EEA

If we start doing that, we should describe other info per country like department, region etc... This way enables to put any sort of address and in addition, this is something already used in other specs (e. g. Seller.json)

> address should be formatted

The purpose of this field is to avoid scenarios such as a german publisher collecting consent for a vendor that operates only in the french market etc... Removing granularity cancels the purpose of this field.
> allow "EU" to use all countries instead of naming them all

janwinkler:
re address formatting: its a different in seller.json and in the cmp. the cmp will need to display it in a readable form. allowing vendors to just add any text, will cause a lot of struggle ...

granulatiry: i dont want to remove DE, FR and so on. just add EU/EEA/EFTA in order to make a shorter list possible

anderagakura:
This is not intended to be disclosed to users in the CMP. It's a standalone file that publisher can use when configuring their CMP to select a subset of vendors rather than the full GVL. 

> re address formatting: its a different in seller.json and in the cmp. the cmp will need to display it in a readable form. allowing vendors to just add any text, will cause a lot of struggle...

Again this is not intended to be disclosed to users in the CMP so there is no need to shorten the list. It's better to have the granularity rather than allowing vendors to select EU by default
> granulatiry: i dont want to remove DE, FR and so on. just add EU/EEA/EFTA in order to make a shorter list possible =>

janwinkler:
re address formating: even if its only for publishers - publishes also want to be able to get data presented in a readable way. plus: what's the downside in formating it?

anderagakura:
There is no downside, we just took this path due to all of our explanations. Of course, agree about getting "_data presented in a readable way_", this is what we want and that's why today we consider this solution as the best one. In the future, if we notice some issues (or you report it) we will review it in order to make some changes. 

> re address formating: even if its only for publishers - publishes also want to be able to get data presented in a readable way. plus: what's the downside in formating it?

HeinzBaumann (approved):

Issue #316, opened 2022-03-14
Clarify use of wildcard identifiers in "Vendor Device Storage & Operational Disclosures"


mattpr:
[Docs](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/Vendor%20Device%20Storage%20%26%20Operational%20Disclosures.md#disclosures-object) indicate wildcards are allowed for `domain`...

> `*` = any domain (ex. first party cookie)
> `*.vendor.com` means multiple subdomains may exist

However it doesn't say whether using wildcards are allowed elsewhere but we have noticed [some vendors](https://help.adspirit.de/deviceStorage.json) that use wildcards in their `identifier` fields (for instance if they have a cookie naming pattern `prefix-customerId`... `prefix-1234` and `prefix-5678` would be covered by `prefix-*`.  

Just wondering whether this is allowed/intended/supported. If so, can we get some clarification in the docs?

dmdabbs:
We will look into it. 
cc: @anderagakura

anderagakura:
@mattpr As mentioned in the [Doc](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/Vendor%20Device%20Storage%20%26%20Operational%20Disclosures.md#disclosures-object), the wildcard is **only** allowed for `domain`. 

Regarding the current status of the doc, if any vendor applies the wildcard outside the `domain` field then it does not follow the doc. Fyi, we have a compliance program in order to check the devicestorage.json of the vendors and notice them if something is wrong. In the meantime, if you see similar issues any devicestorage.json, your feedback are definitely welcome to [framework@iabeurope.eu](mailto:framework@iabeurope.eu)

dmdabbs:
We should investigate AdSpirit's use of wildcards in the name field. I believe Jan Winkler used to work at AdSpirit and could speak to it or connect us to them. 

If a vendor's cookie names incorporate, for example, match partner ids or the like, accommodating the asterisk will be the only tenable way for those vendors or publishers to provide the requested transparency.

dmdabbs:
Jan replied in Slack:

> [Jan Winkler](https://app.slack.com/team/UA34RRYTG)  [6:06 PM](https://iabtechlab-privacy.slack.com/archives/DCP7V4LHE/p1647904008524989)
[...]

PR #320, opened 2022-04-04
Allow wildcards through the field identifier +  adding a new field named domains + disclosures object empty if no client-side storage

anderagakura:


dmdabbs:
Our example for the device disclosure url uses a ".well-known" url  (perhaps from back from when this was required). To avoid people thinking this is required, should we just make it a simple url path?

What about the open question in [Issue 316](https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/issues/316)?

dmdabbs (approved):
LGTM

HeinzBaumann (approved):

spencerwyckoff (approved):
Updates look good and inline with meeting discussions.

PR #328, opened 2022-09-07
Update of the examples and adding a new FAQ

anderagakura:


jdelhommeau:
I am not sure to understand why the "domains" array is set to empty in your example. You seem to insinuate that if a domain is disclosed under "disclosures" section, it does not need to be listed under the "domains" object (that is how I interpret this sentence "The domains array is empty because there is no use of (*) via any domains in the in disclosure array above."). To me, those two sections provide different information and what is disclosed in one shouldn't influence the other.

If you have cookies listed under "disclosures" section, associated to some domains, those same domains should be listed under the "domains" array, which is not the case in your example.

you also changed the example for "Below is sample JSON for a fictional TCF Vendor that does not make use of any client-side storage." to have the "domains" array empty.

"domains" array is not the list of domains used for client-side storage. As per spec, the "domains" array contains all domains that may be used to process personal data, regardless of whether that domain make use of client-side storage. For example, if a vendor process the IP, but doesn't use cookies, then the domain used to call the vendor should be listed under the "domains" section.

anderagakura:
@jdelhommeau The idea was to highlight this scenario but definitely it can be confusing. Going to update it.

> I am not sure to understand why the "domains" array is set to empty in your example. You seem to insinuate that if a domain is disclosed under "disclosures" section, it does not need to be listed under the "domains" object (that is how I interpret this sentence "The domains array is empty because there is no use of (*) via any domains in the in disclosure array above."). To me, those two sections provide different information and what is disclosed in one shouldn't influence the other.

>If you have cookies listed under "disclosures" section, associated to some domains, those same domains should be listed under the "domains" array, which is not the case in your example.

According to what you say, this describes a new example to be added in the doc or in the FAQ. Need to be discussed in mails or in the next meeting.
>you also changed the example for "Below is sample JSON for a fictional TCF Vendor that does not make use of any client-side storage." to have the "domains" array empty.

>"domains" array is not the list of domains used for client-side storage. As per spec, the "domains" array contains all domains that may be used to process personal data, regardless of whether that domain make use of client-side storage. For example, if a vendor process the IP, but doesn't use cookies, then the domain used to call the vendor should be listed under the "domains" section.

jdelhommeau:
actually, amending my comment above. 
`If you have cookies listed under "disclosures" section, associated to some domains, those same domains should be listed under the "domains" array, which is not the case in your example.`
This is only valid if data in cookies is personal, or that processing cookies result in processing personal data (likely the case if cookies are used server side, so along with http request and IP address).

jdelhommeau:
Lets discuss this PR next week during the FSWG meeting as it is unclear to me what you are trying to achieve with those example. From my point of view, those example are wrong, but maybe I am missing something.

anderagakura:
@jdelhommeau They are not wrong. The idea is to highlight, focus the different examples because some vendors just copy/paste without changing according to their mechanism. Anyway, let's discuss about it next week.

janwinkler:
the example says "domains", the table above the example says "domain" ...

jdelhommeau:
@janwinkler the table above list both "domain" and "domains", but there is a note to indicate that "domain" may eventually be deprecated, which is why we are updating examples to use "domains" instead of "domain".

janwinkler:
ah yes i was looking at the old version

PR #329, opened 2022-11-09
Use of the protocol HTTPS over HTTP

anderagakura:


anderagakura:
@jdelhommeau @dmdabbs Now, only HTTPS is considered

PR #345, opened 2023-09-19

viktor-yoc:
Default value for getting int in mPreferences.getInt() cannot be String, it is compilation error. In your case empty string "". It should be Integer and need to be set to 0.


ANGIL21 (approved):

ANGIL21:
perfect

HeinzBaumann (approved):

PR #355, opened 2023-12-04
Fix probably typo on IAB Tech Lab - Consent string and vendor list formats v2.md

peczenyj:
I believe there is a copy paste error on the created field, it has the same description as last updated


HeinzBaumann (changes requested):
The text is correct as is, not a typo. The created and updated timestamp must be identical as of v2.2. - Change Declined. This PR should be closed.

PR #356, opened 2024-01-03
CTV Support

lamrowena:
- Added details for CTV Support


maryxlw (approved):

nicshub (approved):

HeinzBaumann (approved):

janwinkler:
On that last example (Roku): The example uses two different section names (TCF and GPP in ec = CreateObject("roRegistrySection", "TCF")). Im not sure what the implication is, but like to avoid that this breaks something if we dont explicitly call it out in the spec.

Tihana08 (approved):

PR #370, opened 2024-03-25

mujuonly:
Text correction


HeinzBaumann (approved):
This fixes the bookmark in this paragraph. Looks okay.

PR #380, opened 2024-09-07
minor updates

HeinzBaumann:


jdelhommeau:
As discussed during yesterday's FSWG meeting, while we all agreed with the goal, the sentence in its current form created some confusion. We should add further context to it to avoid this once released.
I suggest the below:

> In case the user deselects a vendor (neither giving it consent nor allowing it to operate under legitimate interest) the publisher restriction for this vendor does not need to be written out into the TC string. This is to avoid adding extra restrictions that do not impact the outcome. In case a vendor was not given consent nor legitimate interest, no restriction on that vendor would apply.

PR #404, opened 2025-03-31
Updates to reflect the change in the disclosed vendors section.

HeinzBaumann:
Updated based on the decision made by the TCF FSWG to repurpose the disclosed vendors section to indicate the disclosed vendors and making this section a mandatory section in the TCString.  (see https://docs.google.com/document/d/1evetwMlA_TCJ-GJp3fOqm1k_4W6l6qXA/edit).

### DIFF
--- a/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
+++ b/TCFv2/IAB Tech Lab - Consent string and vendor list formats v2.md
@@ -64,6 +64,7 @@
 
 | Date | Version | Comments |
 | :-- | :-- | :-- |
+| Apr 2025 | 2.3 | Resolve the LI ambiguity by repurposing and making the ‘Disclosed Vendors’ section a mandatory section of the TC string.
 | Feb 2024 | 2.2 | Deprecated the gdpr_pd macro, and added the field 'environments' to the CMP JSON. |
 | May 2023 | 2.2 | Update to further strengthen the TCF as a standard in the industry: revised purpose names and descriptions, introduced retention periods for all purposes, removed legitimate interest for purposes 3 to 6, the introduction of data categories used in conjunction with the purposes, support for legitimate interest claim urls, adding support for localized policy urls and introducing a more robust vendor compliance program. |
 | June 2022 | 2.1 | Update of the <b>Global Vendor List JSON Object</b> example regarding the filename in `deviceStorageDisclosureUrl` |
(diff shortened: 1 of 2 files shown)


janwinkler (approved):

HeinzBaumann:
I will need to update the TC String example with a disclose vendors segment in the segment section of the spec.

patmmccann:
It is currently impossible for a user to register an objection to SP1, and several vendors, including Google Ad Manager (in the __eoi) cookie disclosed here ( https://business.safety.google/adscookies/ ) but not in the device disclosure url in the tcf registration) and the Ozone Project (disclosed here https://github.com/prebid/prebid.github.io/pull/5967/files ), seem to rely on Special Purpose 1 to set cookies. Google Ad Manager documentation says there is nothing in tcdata that can prevent this cookie from being set by pubads_impl (https://support.google.com/admanager/answer/9882911?hl=en#limited-ads-update), as they only rely on special purposes to run Limited Ads 2.0.

Since there is nothing allowing an SP1 objection in TCData, we have no way to communicate to these vendors this is unacceptable for this user. 

We'd like for an SP objection to be specifically possible to express in tcdata, as the e-privacy directive does not seeem to allow for device access for services not specifically requested by the user, and some publishers may need to register or express special purpose objections to prevent vendors with unusual legal interpretations from accessing the device.

achimschloss:
This seems to need an alignment in the TCF policy working group - at least with Ozone they argue that (quoting the PR) _"Consent: As this cookie is strictly necessary for the functioning and security of the website, it is exempt from consent requirements under applicable data protection laws, including the UK GDPR and ePrivacy Directive."_, I'd assume likewise with Google.
 
In that sense they do not rely on SP1 to place the cookie (which they also cannot based on the policy), but requesting controls for SP1 is a topic that can be (re)discussed.

jdelhommeau:
thank you @patmmccann  for your feedback. While I understand the need and the requirement, I do not believe they are directly linked with the current PR, which only goals is to remove existing ambiguity. As such, we are moving forward with the current PR.

We had however already explored the ability for publishers to perform publisher restrictions, including on Special Purposes, which is what I believe you are after.
Your concern will be raised again in the Policy Working group and if the group is supportive, the Framework Signal Working Group will work with the tech lab to update the TCF specs accordingly.

HeinzBaumann:
Update based on FSWG input. Added support for disclosed vendors in mobile storage, added a Q&A to the document for this change and update the broken links in the guideline document.

PR #411, opened 2025-07-29
Try to make AVI specification match reality

jaylett-annalect:
Missing possible values, differences in spelling, and differences in capitalisation between the spec and the live file.


HeinzBaumann:
@lamrowena can you please review this quickly. Thanks!

PR #426, opened 2025-11-22
added clarification for segment order

HeinzBaumann:
Based on TCF FSWG discussion I added a clarification on the order of the segments in the tcString.


jdelhommeau:
Thank you @HeinzBaumann . All good for me. Unfortunately,  i don't sem to have ability to review and validate the PR... @lamrowena can you grant me access to review?

PR #437, opened 2026-03-04
Remove SP-only vendor workaround from spec

HeinzBaumann:
removed workaround from the specification
update the history


jdelhommeau:
thank you @HeinzBaumann ! I believe during the wg, we agreed to also add some text about it in the question in first part of the specs. Shouldn't we ship both changes together?

HeinzBaumann:
@jdelhommeau it is change line 340.

HeinzBaumann:
@lamrowena This is ready to be merged

lamrowena (approved):

PR #445, opened 2026-06-10
July 2026 vendor disclosure spec update

HeinzBaumann:
Clarified the declaration of domains
Added purposes and special purposes to domain and sdk sections
Updated examples
Fixed styling issues

### DIFF
--- a/TCFv2/Vendor Device Storage & Operational Disclosures.md
+++ b/TCFv2/Vendor Device Storage & Operational Disclosures.md
@@ -180,7 +180,9 @@ Below is sample JSON for a fictional TCF Vendor that does not make use of any `c
 
 ### Domains array
 
-Vendors MUST publish the domains they use for collecting and processing personal data in the context of their TCF registration. This may include domains used for the delivery of assets, such as image hosting URLs, or ad delivery servers where a client-side network call necessarily involves the automatic transmission of information that can be considered personal data under the GDPR.
+Vendors MUST publish the domains they use for collecting and processing personal data in the context of their TCF registration. Vendors MUST publish the domains they use for collecting and processing personal data in the context of their TCF registration. This may include domains that are not used for storing and/or accessing information on users' devices (e.g. domains used for the delivery of assets, such as image hosting URLs, or ad delivery servers where a client-side network call necessarily involves the automatic transmission of information that can be considered personal data under the GDPR).
+Vendors MUST NOT include Publishers’ delegated domains or subdomains they may use.
+
 Vendors MUST NOT include Publishers’ delegated domains or subdomains they may use.
 
 <table>
--- a/TCFv2/Vendor Device Storage & Operational Disclosures.md
+++ b/TCFv2/Vendor Device Storage & Operational Disclosures.md
@@ -269,6 +269,7 @@ Vendors that support mobile in-app environments but do not use any SDKs MUST inc
 <br>If no SDKs are used, an empty array must be included.
 <br>If the vendor does not operate in mobile in-app environments, the SDKs array may be omitted.
 </td></tr>
+ <tr><td><code>name</code></td><td>required</td><td>string</td><td>The API package name as registered with the corresponding mobile app store.</td></tr>
 <tr>
    <td><code>purposes</code>
    </td>
--- a/TCFv2/Vendor Device Storage & Operational Disclosures.md
+++ b/TCFv2/Vendor Device Storage & Operational Disclosures.md
@@ -2,7 +2,7 @@
 
  **IAB Europe Transparency & Consent Framework**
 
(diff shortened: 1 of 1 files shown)


TCFv2/Vendor Device Storage & Operational Disclosures.md
```diff
@@ -182,7 +183,8 @@ Below is sample JSON for a fictional TCF Vendor that does not make use of any `c
 
 ### Domains array
 
-Vendors MUST publish the domains they use to collect and process personal data in the context of their TCF registration. Vendors MUST NOT include Publishers’ delegated domains or subdomains that they may operate on behalf of a Publisher. Vendors that do not use domains (i.e., no domains are declared in the disclosures array) or that do not operate in a Web environment SHOULD NOT include the domains array.
+Vendors MUST publish the domains they use for collecting and processing personal data in the context of their TCF registration. This may include domains that are not used for storing and/or accessing information on users' devices (e.g. domains used for the delivery of assets, such as image hosting URLs, or ad delivery servers where a client-side network call necessarily involves the automatic transmission of information that can be considered personal data under the GDPR).
```
simontrasler, line 186: The following point got removed: is that intentional?
```
Vendors that do not use domains (i.e., no domains are declared in the disclosures array) or that do not operate in a Web environment SHOULD NOT include the domains array.
```
FWIW I don't understand the point the original text was trying to make, so it may be better to delete it. Typically, an empty array would map to a vendor that uses no domains, and an omitted array would map to "don't know".

TCFv2/Vendor Device Storage & Operational Disclosures.md
```diff
@@ -195,6 +197,26 @@ Vendors MUST publish the domains they use to collect and process personal data i
 <br><br>
 Entry MUST NOT contain “http(s)://” or text other than the domain.
 </td></tr>
+<tr>
+   <td><code>purposes</code>
+   </td>
+   <td>required
+   </td>
+   <td>array&lt;integer>
```
simontrasler: Nit: this terminology may be confusing. I'd recommend following the style used in the OpenRTB spec, e.g., "array of integers".

TCFv2/Vendor Device Storage & Operational Disclosures.md
```diff
@@ -195,6 +197,26 @@ Vendors MUST publish the domains they use to collect and process personal data i
 <br><br>
 Entry MUST NOT contain “http(s)://” or text other than the domain.
 </td></tr>
+<tr>
+   <td><code>purposes</code>
+   </td>
+   <td>required
```
simontrasler, line 203: I do not expect we can add new `required` fields without an accompanying version bump: is that being done? Or the new fields can be made `optional`.

lamrowena: This is actively being discussed :)

live4pedro-ux:
Hi Simon,

Regarding the points raised on PR #445 for the July 2026 vendor disclosure
spec update:

The removal of the text concerning the domains array was intentional, as we
agree the original wording was confusing. We will proceed with deleting it
from the specification.

Thank you for the note regarding the terminology. I will update
"array<integer>" to "array of integers" to align with the OpenRTB spec
style.

Regarding the version bump for the new required fields, this is actively
being discussed internally, and we will confirm our versioning approach
with you soon.

Best regards,

Fernando Rivera

TCFv2/Vendor Device Storage & Operational Disclosures.md
HeinzBaumann: This was intentially revised based on TCF's FSWG feedback

TCFv2/Vendor Device Storage & Operational Disclosures.md
HeinzBaumann: Yes, we can bump the version to 1.2.

TCFv2/Vendor Device Storage & Operational Disclosures.md
HeinzBaumann: We can change this.

TCFv2/Vendor Device Storage & Operational Disclosures.md
```diff
@@ -2,7 +2,7 @@
 
  **IAB Europe Transparency & Consent Framework**
 
-**Final v.2.0 | August 2019, Updated November 2022, December 2025**
+**Final v.2.0 | August 2019, Updated November 2022, December 2025, July 2026**
```
HeinzBaumann: @lamrowena How did we end up with v.2.0 when in the history we have 1.x? Do you remember?

live4pedro-ux (approved):
https://accounts.google.com/gsi/transform

PR #451, opened 2026-08-11
Update Vendor Device Storage & Operational Disclosures.md

Lucio-IAB:
Link to FAQ's PDF is outdated

### DIFF
--- a/TCFv2/Vendor Device Storage & Operational Disclosures.md
+++ b/TCFv2/Vendor Device Storage & Operational Disclosures.md
@@ -298,4 +298,4 @@ For secure communications, the vendor must make publicly accessible the JSON fil
 
 ## FAQ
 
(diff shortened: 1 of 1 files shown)


Lucio-IAB:
or else, we should keep the same URL (https://iabeurope.eu/transparency-consent-framework-file/tcf-framework-faq-vendor-device-storage-and-operational-disclosures.pdf) and overwrite the old PDF with the new one (https://iabeurope.eu/wp-content/uploads/260610-FAQs-on-Vendor-Device-Storage-Operational-Disclosures.pdf)

HeinzBaumann:
@Lucio-IAB We should keep the old URL.

Lucio-IAB:
> @Lucio-IAB We should keep the old URL.

Agreed. I just received FTP access, so I should be able to do that. Thanks!

